PR #29, opened 2021-03-31
Add capacity reservation support

shin-nien:

This PR allows machines to target an [AWS Resource Group containing Capacity Reservations](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-using.html#create-cr-group).

Sometimes AWS is unable to create new instances due to a temporary shortage of capacity in an AZ. AWS Capacity Reservations minimise the risk of stalling deployments; especially ones that use a large maxSurge (which creates more instances than you're currently running).

The default reservation target type is `open` so you can create the reservation group ahead of time without any reservations in it. If machines are created and can't match any reservations then they're created using On-Demand capacity (like the usual way).

Typically, you'd create an AWS Reservation Group ahead of time with capacity reservations inside it; normally you'd create a reservation of `replicas + maxSurge` instances for a particular platform, instance type and AZ. Any instances already running will immediately make use of the reservation but you will then have a buffer to guarantee a rolling deployment.

Targeting a reservation group is preferred because it is AZ agnostic just like Machine Classes are. So it doesn't make much sense to want to target a capacity reservation directly (instead of a group) but the option is included in this PR just in case. 

- I wasn't sure if I needed to do anything with migration.go as I as adding new fields so any existing classes wouldn't have them defined, any advice would be appreciated!

**Release note**:

```improvement operator
Support for targeting an AWS Resource Group containing capacity reservations - https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-using.html#create-cr-group
```

### DIFF
--- a/kubernetes/machine-class.yaml
+++ b/kubernetes/machine-class.yaml
@@ -42,6 +42,9 @@ providerSpec:
 secretRef: # Secret pointing to a secret which contains the provider secret and cloudconfig
   namespace: default  # Namespace
   name: test-secret # Name of the secret
+# capacityReservation: # Optional - allows targeting of an AWS Capacity reservation or an AWS Resource Group containing reservations. Use either a reservation id or group arn - but not both.
+#   capacityReservationId: "cr-05c28b843c05acc11"
+#   capacityReservationResourceGroupArn: "arn:aws:resource-groups:us-west-1:123456789012:group/my-cr-group"
 # credentialsSecretRef: # Optional - Kubernetes secret containing only provider secrets (in this case the Secret in the secretRef does not need them)
 #   name: "test-secret-credentials" # Name of the secret
 #   namespace: "default" # Namespace of secret
--- a/kubernetes/machine-class.yaml
+++ b/kubernetes/machine-class.yaml
@@ -24,6 +24,9 @@ providerSpec:
         volumeType: gp2 # Type of the root block device
         encrypted: true
         deleteOnTermination: true
+#  capacityReservation: # Optional - allows targeting of an AWS Capacity reservation or an AWS Resource Group containing reservations. Only one needs to be specified.
+#   capacityReservationId: "cr-05c28b843c05acc11"
+#   capacityReservationResourceGroupArn: "arn:aws:resource-groups:us-west-1:123456789012:group/my-cr-group"
   iam:
     name: iam-name # Name of the AWS Identity and Access Management
   keyName: key-value-pair-name # EC2 keypair used to access ec2 machine
@@ -42,9 +45,6 @@ providerSpec:
 secretRef: # Secret pointing to a secret which contains the provider secret and cloudconfig
   namespace: default  # Namespace
   name: test-secret # Name of the secret
-# capacityReservation: # Optional - allows targeting of an AWS Capacity reservation or an AWS Resource Group containing reservations. Only one needs to be specified.
-#   capacityReservationId: "cr-05c28b843c05acc11"
-#   capacityReservationResourceGroupArn: "arn:aws:resource-groups:us-west-1:123456789012:group/my-cr-group"
 # credentialsSecretRef: # Optional - Kubernetes secret containing only provider secrets (in this case the Secret in the secretRef does not need them)
 #   name: "test-secret-credentials" # Name of the secret
 #   namespace: "default" # Namespace of secret
--- a/kubernetes/machine-class.yaml
+++ b/kubernetes/machine-class.yaml
@@ -42,7 +42,7 @@ providerSpec:
 secretRef: # Secret pointing to a secret which contains the provider secret and cloudconfig
   namespace: default  # Namespace
   name: test-secret # Name of the secret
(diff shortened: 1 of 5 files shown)


gardener-robot-ci-1:
Thank you @shin-nien for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

kubernetes/machine-class.yaml
```diff
@@ -42,6 +42,9 @@ providerSpec:
 secretRef: # Secret pointing to a secret which contains the provider secret and cloudconfig
   namespace: default  # Namespace
   name: test-secret # Name of the secret
+# capacityReservation: # Optional - allows targeting of an AWS Capacity reservation or an AWS Resource Group containing reservations. Use either a reservation id or group arn - but not both.
```
prashanth26, line 45: If this is an either/or field, it would probably make sense to check this at the time of [validation of the spec here ](https://github.com/gardener/machine-controller-manager-provider-aws/blob/master/pkg/aws/apis/validation/validation.go#L59)right? You can probably have a new helper method to validate the same.

pkg/aws/core_test.go
```diff
@@ -151,6 +151,38 @@ var _ = Describe("MachineServer", func() {
 					errToHaveOccurred: false,
 				},
 			}),
+			Entry("Machine creation request for capacity reservations", &data{
```
prashanth26, line 154:
```suggestion
			Entry("Machine creation request for capacity reservations with capacityReservationId", &data{
```

pkg/aws/core_test.go
```diff
@@ -151,6 +151,38 @@ var _ = Describe("MachineServer", func() {
...
+				},
+				expect: expect{
+					machineResponse: &driver.CreateMachineResponse{
+						ProviderID: "aws:///eu-west-1/i-0123456789-0",
+						NodeName:   "ip-0",
+					},
+					errToHaveOccurred: false,
+				},
+			}),
+			Entry("Machine creation request for an AWS Capacity Reservation Group", &data{
```
prashanth26, line 170:
```suggestion
			Entry("Machine creation request for an AWS Capacity Reservation Group with capacityReservationResourceGroupArn", &data{
```

pkg/aws/core_test.go
```diff
@@ -151,6 +151,38 @@ var _ = Describe("MachineServer", func() {
...
+						Secret:       providerSecret,
+					},
+				},
+				expect: expect{
+					machineResponse: &driver.CreateMachineResponse{
+						ProviderID: "aws:///eu-west-1/i-0123456789-0",
+						NodeName:   "ip-0",
+					},
+					errToHaveOccurred: false,
+				},
```
prashanth26, line 184: It would be great to have an negative test case to check for setting both `capacityReservationResourceGroupArn` and `capacityReservationId` leading to a failure in validation as described above.

kubernetes/machine-class.yaml
shin-nien: Ah yes that makes more sense. Done.

pkg/aws/core_test.go
shin-nien: Done.

prashanth26 (approved):
Cool thanks for the changes. 
/lgtm from my side. @AxiomSamarth can you also have a look at it. Just for a 4 eye principle.

AxiomSamarth (approved):

kubernetes/machine-class.yaml
```diff
@@ -42,6 +42,9 @@ providerSpec:
 secretRef: # Secret pointing to a secret which contains the provider secret and cloudconfig
   namespace: default  # Namespace
   name: test-secret # Name of the secret
+# capacityReservation: # Optional - allows targeting of an AWS Capacity reservation or an AWS Resource Group containing reservations. Only one needs to be specified.
+#   capacityReservationId: "cr-05c28b843c05acc11"
+#   capacityReservationResourceGroupArn: "arn:aws:resource-groups:us-west-1:123456789012:group/my-cr-group"
```
prashanth26, line 47: Hi @shin-nien,

One last change. The sample machine class seems to be wrong here. I think these new fields have to be moved under the `ProviderSpec` [here](https://github.com/gardener/machine-controller-manager-provider-aws/blob/7ca57524d7590afbf1fdfabf72e5205b8db283ec/kubernetes/machine-class.yaml#L12-L41)?

prashanth26 (changes requested):

kubernetes/machine-class.yaml
shin-nien: Ah yes, of course. Thanks.

Done!

prashanth26 (approved):
Thank you for the quick changes.
/lgtm

prashanth26:
@shin-nien - I have fixed some go lint issue on this PR [here](https://github.com/gardener/machine-controller-manager-provider-aws/pull/33). I hope the change is okay as is retains the [lowerCase](https://github.com/gardener/machine-controller-manager-provider-aws/pull/33/files#diff-d30d0813f949736a1f919d711ad36e6c923f8199b135fcefd2fb0125ad10208aR136) in the actual spec definition.

PR #32, opened 2021-04-20
introduced logic to set "Placement" field in machine using annotations

### LINKED ISSUE #24: Support to provision dedicated host at AWS
We have the need to start instances on dedicated hosts at AWS. Some of the huge memory instances like u-6tb1.metal are only available through dedicated hosts which are exclusively assigned into a customers account at AWS.

To create instances on these dedicated host, the AWS ec run-instance command needs to support the placement parameter.

Example command:
aws ec2 run-instances --image-id ami-xxxx --instance-type u-6tb1.metal --placement Affinity=host,Tenancy=host,AvailabilityZone=eu-central-1b --security-group-ids sg-abcd --subnet-id subnet-xyz --region eu-central-1

himanshu-kun:
To enable users to run VMs on a dedicated host on their account. Annotations would have to set by them in the worker in shoot spec to specify tenancy, affinity of their placement and availability zone of their dedicated host , assuming auto-placement is enabled on the host. 

Fixes #24 

I have added a variadic parameter name params to `newMachine(--)` in pkg/aws/aws_suite_test.go along with comment on whats it for. 

**Release note**:

```noteworthy user
Allow setting of `placement` parameters on AWS EC2 instances using a specific key on machine annotations. For setting the annotation refer to the doc provided [here](https://github.com/gardener/machine-controller-manager-provider-aws/blob/5cd832aadd0e5eb3feef56186e004fb7152c55f2/kubernetes/machine-deployment.yaml#L35-L38). 
```


gardener-robot-ci-1:
Thank you @himanshu-kun for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/aws/core.go
```diff
@@ -171,6 +176,38 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
 }
 
+//returns Placement Object required in ec2.RunInstancesInput
+func getPlacementObj(req *driver.CreateMachineRequest) *ec2.Placement {
+
+	tenancy := "default"
+	affinity := "default"
+	var availZone string
+
+	myannotations := req.Machine.Spec.NodeTemplateSpec.ObjectMeta.Annotations
```
AxiomSamarth, line 186: Nit picking - How about using a better variable name instead of `myannotations`?

pkg/aws/core.go
```diff
@@ -171,6 +176,38 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 	return response, nil
 }
 
+//returns Placement Object required in ec2.RunInstancesInput
+func getPlacementObj(req *driver.CreateMachineRequest) *ec2.Placement {
+
+	tenancy := "default"
+	affinity := "default"
```
AxiomSamarth, line 183: Why not use `ec2.AffinityDefault` and `ec2.TenancyDefault` instead of default?

pkg/aws/core.go
```diff
@@ -171,6 +176,38 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+	for key, value := range myannotations {
+
+		if key == awsPlacementAffinity && value != "" {
+			affinity = value
+		} else if key == awsPlacementAvailabilityZone && value != "" {
+			availZone = value
+		} else if key == awsPlacementTenancy && value != "" {
+			tenancy = value
+		}
+	}
```
AxiomSamarth, line 197: If these three keys (`awsPlacementAffinity`, `awsPlacementAvailabilityZone`, `awsPlacementTenancy`) are the only keys of interest then we do not have to iterate over all the annotations. Considering this is a map type, the emptiness of the value can be check and directly assigned.

pkg/aws/aws_suite_test.go
```diff
@@ -32,15 +32,18 @@ const (
 	testNamespace = "test"
 )
 
+//params is used to pass annotations to the machine spec
```
AxiomSamarth, line 35: Let's update the comment as

```
params is the variadic parameter of type interface{} for future adaption of the newMachine function. 
The first of the variadic params is reserved for Machine.Spec.NodeTemplateSpec.ObjectMeta.Annotations
``` 
so that it is clear for other developers in futures

pkg/aws/core.go
prashanth26: Nice catch @AxiomSamarth . I think this suggestion makes sense.

pkg/aws/core_test.go
```diff
@@ -39,6 +39,11 @@ const (
 	cloudProviderReturnedError            = "machine codes error: code = [Internal] message = [Cloud provider returned error]"
 )
 
+var (
+	affinity_const  = ec2.AffinityDefault
+	tenancy_const   = ec2.TenancyDefault
+	availZone_const = "eu-west-1a"
```
prashanth26, line 45:
```suggestion
	affinityConst  = ec2.AffinityDefault
	tenancyConst   = ec2.TenancyDefault
	availZoneConst = "eu-west-1a"
```

GoLang generally prefers the approach of mixedCases. Refer - https://golang.org/doc/effective_go#mixed-caps. Also, why not use `ec2.AffinityDefault` and `ec2.TenancyDefault` directly? Also please adopt the rest of the code with the same.

pkg/aws/core.go
```diff
@@ -171,6 +176,38 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 	return response, nil
 }
 
+//returns Placement Object required in ec2.RunInstancesInput
+func getPlacementObj(req *driver.CreateMachineRequest) *ec2.Placement {
```
prashanth26, line 180: From what i see, I understand that the method returns an placement object with tenancy and affinity set as `Default` in the normal case. Just wondering what this maps to in the AWS context? Could it be the case that AWS doesn't understand the placement object if there are no dedicated hosts in the account? We will need to cross check this with the documentation. 

An alternative approach could be to return a `nil` object when nothing is set. And to handle this nil object to not add this entry in the final creation request. This would let the current code as it is today and only add the placement object if anyone of the fields are set.

pkg/aws/core.go
```diff
@@ -141,6 +145,7 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		},
 		NetworkInterfaces: networkInterfaceSpecs,
 		TagSpecifications: []*ec2.TagSpecification{tagInstance, tagVolume},
+		Placement:         getPlacementObj(req),
 	}
```
prashanth26, line 150:
```suggestion
		NetworkInterfaces: networkInterfaceSpecs,
		TagSpecifications: []*ec2.TagSpecification{tagInstance, tagVolume},
	}
	if placement := getPlacementObj(req); placement != nil {
	  inputConfig.Placement =  placement
	}
```
If you take my suggestion, the code would look something like this. This way current code/requests are not affected.

pkg/aws/core_test.go
```diff
@@ -50,6 +55,11 @@ var _ = Describe("MachineServer", func() {
 			"userData":                []byte("dummy-user-data"),
 		},
 	}
+	annotations := map[string]string{
+		awsPlacementAffinity:         affinity_const,
+		awsPlacementAvailabilityZone: availZone_const,
+		awsPlacementTenancy:          tenancy_const,
```
prashanth26, line 61: I see that you have test cases with `Default` as a value. Would be nice to have a couple of test cases with customer requirement values of `host`,`host`.

pkg/aws/core_test.go
```diff
@@ -508,6 +626,7 @@ var _ = Describe("MachineServer", func() {
 				ms := NewAWSDriver(mockPluginSPIImpl)
 				ctx := context.Background()
 
+				//if there is a create machine request by the test case then create the machine
```
prashanth26, line 629:
```suggestion
				// If there is a create machine request by the test case then create the machine
```

pkg/aws/aws_suite_test.go
```diff
@@ -32,15 +32,18 @@ const (
...
 func newMachine(
-	setMachineIndex int,
+	setMachineIndex int, params ...interface{},
```
prashanth26, line 37:
```suggestion
	setMachineIndex int, annotations map[string]string,
```
I would have expected this method to be simpler like this. This would help future contributors to easily understand the expected parameters. I know the interface would provide the flexibility to add any type/number of params, however, I feel it would be difficult for new contributors to understand. 

However, if you have strong arguments for having interfaces please let me know as I am not an expert myself and may have overlooked the advantages.

prashanth26 (commented):
Thanks for this @himanshu-kun . Just minor nits. Looks good to me overall.

pkg/aws/aws_suite_test.go
AxiomSamarth: @prashanth26 Actually, this was my idea/suggestion to Himanshu to introduce the variadic parameter of type empty interface exactly for the reason as you have inferred `the interface would provide the flexibility to add any type/number of params`. So probably, I will explain the reasoning.

The reason I thought of this approach is that any new parameter addition to the function `newMachine()` by a developer in future will not force the developer to modify all the function calls at other places.

However, I missed thinking about the fact that you mentioned
>I feel it would be difficult for new contributors to understand

I second your suggestion here 😄 Thank you!

prashanth26: > The reason I thought of this approach is that any new parameter addition to the function newMachine() by a developer in future will not force the developer to modify all the function calls at other places.

I agree with you as well. I thought about this after I made my comment. In some sense, I guess both have it's advantages and disadvantages. 

Himanshu - If you feel the change to annotations as suggested above it's too tiring/difficult to make. I think a simple `ctrl+f` replace `newMachine(<regexp>, <number>)` to `newMachine(<regexp>, <number>, nil)` should do the trick to fix other invokations to this method. However, If you find it difficult I think the current approach is also okay. However, we should improve the comments with @AxiomSamarth 's suggestion above so that new contributors aren't confused. :)

AxiomSamarth: I just looked up for some best practices with respect to using empty interfaces and below are the highlights/findings-

- `interfaces{}` should be the last thing to be used when there is no other straight forward way because using `interface{}` would also carry a burden of introducing type assertions which it already has [here](https://github.com/himanshu-kun/machine-controller-manager-provider-aws/blob/9e8f0cdf154c50bd7f80ad427894c01d852d3788/pkg/aws/aws_suite_test.go#L61) (which in turn may carry some programmatic limitations).

In this situation, introducing apt parameter of type `map[string]string` looks like a better option to me. Again, even I shall leave it to Himashu's decision 😄

prashanth26: > I just looked up for some best practices with respect to using empty interfaces and below are the highlights/findings-
> 
> * `interfaces{}` should be the last thing to be used when there is no other straight forward way because using `interface{}` would also carry a burden of introducing type assertions which it already has [here](https://github.com/himanshu-kun/machine-controller-manager-provider-aws/blob/9e8f0cdf154c50bd7f80ad427894c01d852d3788/pkg/aws/aws_suite_test.go#L61) (which in turn may carry some programmatic limitations).
> 
> In this situation, introducing apt parameter of type `map[string]string` looks like a better option to me. Again, even I shall leave it to Himashu's decision 😄

I would also prefer the `map[string]string` param approach. I would think it's less confusing to new developers. Let's do with this then.

pkg/aws/core_test.go
himanshu-kun: I used the constants so that we could change ec2.AffinityDefault to lets say ec2.AffinityHost at just one placeand in the expect block of one of the test cases , since we have to initialzie &value, using of var affinityConst and tenancyConst seemed the only option to me, unless there's a way to create variable on the fly and assign their address to a field.

pkg/aws/aws_suite_test.go
himanshu-kun: I am thinking of changing `newMachine(setMachineIndex int, params ...interface{})` to `newMachine(setMachineIndex int, params ...map[string]string)`
also I didn't get the `ctrl+f` approach mentioned by @prashanth26

prashanth26: >I am thinking of changing newMachine(setMachineIndex int, params ...interface{}) to newMachine(setMachineIndex int, params ...map[string]string)

Yes, that is the only change required.

> I didn't get the ctrl+f approach mentioned by @prashanth26

This was just a trick for you to update existing method invocations to the new signature style.

pkg/aws/core.go
```diff
@@ -171,6 +178,44 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+		placementobj.AvailabilityZone = &availZone
+	}
+
+	if *placementobj == (ec2.Placement{}) {
+		return nil
+	}
+
+	return placementobj
+}
+
```
AxiomSamarth, line 218: I see there is a scope to write the `getPlacementObj` function in a cleaner way. Recommending here. Just see if we can adopt this way

```suggestion

//returns Placement Object required in ec2.RunInstancesInput
func getPlacementObj(req *driver.CreateMachineRequest) *ec2.Placement {

	var placementobj = &ec2.Placement{}

	requestAnnotations := req.Machine.Spec.NodeTemplateSpec.ObjectMeta.Annotations

	if affinity, ok := requestAnnotations[awsPlacementAffinity]; ok && affinity != "" {
		placementobj.Affinity = &affinity
	}
	if availZone, ok := requestAnnotations[awsPlacementAvailabilityZone]; ok && availZone != "" {
		placementobj.AvailabilityZone = &availZone
	}
	if tenancy, ok := requestAnnotations[awsPlacementTenancy]; ok && tenancy != "" {
		placementobj.Tenancy = &tenancy
	}

	if *placementobj == (ec2.Placement{}) {
		return nil
	}

	return placementobj
}

```

pkg/aws/core_test.go
```diff
@@ -289,6 +317,96 @@ var _ = Describe("MachineServer", func() {
 		)
 	})
 
+	Describe("#GetPlacementObject", func() {
```
AxiomSamarth, line 320: Can we also add one positive test case in this?

pkg/aws/core.go
himanshu-kun: yes sure

prashanth26 (commented):
Looks like this branch is behind master. You will have to rebase the changes with master. However, do backup your current branch before you do it to avoid losing any changes while rebasing. 

```
git checkout hana-annotations
git checkout -b hana-annotations-bk

git checkout master
git pull origin master (assuming origin to point to gardener)
git checkout hana-annotations
git rebase master

File: pkg/aws/core.go [hunk 13]
```diff
@@ -42,7 +42,10 @@ const (
 	resourceTypeInstance = "instance"
 	resourceTypeVolume   = "volume"
 	// awsEBSDriverName is the name of the CSI driver for EBS
-	awsEBSDriverName = "ebs.csi.aws.com"
+	awsEBSDriverName                    = "ebs.csi.aws.com"
+	awsPlacementAffinity         string = "machine.sapcloud.io/awsPlacementAffinity"
+	awsPlacementAvailabilityZone string = "machine.sapcloud.io/awsPlacementAvailabilityZone"
+	awsPlacementTenancy          string = "machine.sapcloud.io/awsPlacementTenancy"
```
prashanth26, line 48:
```suggestion
	awsPlacementAffinity                = "machine.sapcloud.io/awsPlacementAffinity"
	awsPlacementAvailabilityZone = "machine.sapcloud.io/awsPlacementAvailabilityZone"
	awsPlacementTenancy              = "machine.sapcloud.io/awsPlacementTenancy"
```
String is implicit, we can ignore it.

prashanth26 (commented):
I hope you have tested these changes with an actual shoot?

prashanth26 (commented):
Can we also add an example so that end users can take advantage of this feature looking at the example? 
We will have to update these files - https://github.com/gardener/machine-controller-manager-provider-aws/tree/master/kubernetes

pkg/aws/core.go
himanshu-kun: made these changes, thanks for the suggestion

pkg/aws/core_test.go
himanshu-kun: done, thanks

himanshu-kun:
> I hope you have tested these changes with an actual shoot?

I will t

himanshu-kun:
> I hope you have tested these changes with an actual shoot?

yes I have tested it and its running as expected

himanshu-kun:
> Can we also add an example so that end users can take advantage of this feature looking at the example?
> We will have to update these files - https://github.com/gardener/machine-controller-manager-provider-aws/tree/master/kubernetes

I have added the example in the specified files

AxiomSamarth (approved):

kubernetes/machine-deployment.yaml
```diff
@@ -31,4 +31,10 @@ spec:
...
         name: test-mc # Name of the machine class
-        namespace: default
\ No newline at end of file
+        namespace: default
+      # nodeTemplate: 
+      #   metadata:
+      #     annotations: # Optional: allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled 
+      #       machine.sapcloud.io/awsPlacementAffinity: host #Optional: can't take host unless Tenancy is also host
+      #       machine.sapcloud.io/awsPlacementAvailabilityZone: eu-west-1a #Optional
+      #       machine.sapcloud.io/awsPlacementTenancy: host #Optional
```
prashanth26, line 40: This makes the `end-user` think that the `nodeTemplate` is a subsection of `class` but when in reaility it's a subsection of `spec` as seen from [this](https://github.com/gardener/machine-controller-manager/blob/master/pkg/apis/machine/v1alpha1/machine_types.go#L63). Can we keep the comments before the alignment? Refer https://github.com/gardener/gardener/blob/master/example/90-shoot.yaml#L49-L67. 

It would be great if you can move the other optional fields similarily [here](https://github.com/gardener/machine-controller-manager-provider-aws/blob/f81603460c30057c71426cab9f53ad1080512db6/kubernetes/machine-deployment.yaml#L13-L15)

kubernetes/machine.yaml
```diff
@@ -14,3 +14,9 @@ spec:
     kind: MachineClass
     name: test-mc
     namespace: default
+  # nodeTemplate: 
+  #   metadata:
+  #     annotations: # Optional: allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled 
+  #       machine.sapcloud.io/awsPlacementAffinity: host #Optional: can't take host unless Tenancy is also host
+  #       machine.sapcloud.io/awsPlacementAvailabilityZone: eu-west-1a #Optional
+  #       machine.sapcloud.io/awsPlacementTenancy: host #Optional
```
prashanth26, line 22: Same as above

prashanth26 (commented):
Couple of last minor changes.

pkg/aws/core.go
himanshu-kun: @prashanth26  I had used the `commit suggestion` button here to commit this change you suggested and new commit was made(commit 1)..after that I pushed my commit with other changes from local repo but there I hadn't made these suggested changes(commit2)
I thought the two commits will be different and both will be present here, but commit1 was removed and commit2 remained... I would be happy to know how did it happen
Anyways I have made the changes and will make a new commit with all of them

pkg/aws/core_test.go
```diff
@@ -289,6 +301,22 @@ var _ = Describe("MachineServer", func() {
 					errMessage:        "machine codes error: code = [Internal] message = [Region doesn't exist while trying to create session]",
 				},
 			}),
+			Entry("Placement object with all fields set", &data{
```
amshuman-kr, line 304: Is there a way to check if the placement group was set on the AWS machine create request?

pkg/aws/aws_suite_test.go
```diff
@@ -32,15 +32,18 @@ const (
 	testNamespace = "test"
 )
 
+//params is used to pass annotations to the machine spec
 func newMachine(
-	setMachineIndex int,
+	setMachineIndex int, params ...map[string]string,
```
amshuman-kr, line 37: Why variadic? I see only `params[0]` being used [below](https://github.com/gardener/machine-controller-manager-provider-aws/pull/32/files#diff-22e0e177995875730b9c602d1be2aefcdd624ddec10f8be8b5e31f606bc20925R71).

Why not just `	setMachineIndex int, annotations map[string]string,`?

amshuman-kr (changes requested):
Thanks for the PR @himanshu-kun! The changes overall look good. Just small queries about the tests below.

Also, the approach seems to be midway between the single boolean annotations and full support. Any particular reason for supporting multiple annotations? I ask this because the annotations as they are right now still support only a subset of the [AWS placement group](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Placement.html) functionality. If you want to support variations in placement group configuration, an alternative could be to support a single annotation key with a JSON serialized [placement object](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Placement.html) as a value.

E.g.
```yaml
annotations:
  machine.sapcloud.io/awsPlacement: '{ "Affinity": "host", "AvailabilityZone", "eu-west-1a", "Tenancy": "host"}'
```

The advantage is single annotation for placement and also you can seamlessly support other possible placement configurations. WDYT?

pkg/aws/aws_suite_test.go
himanshu-kun: Yes @amshuman-kr  I had thought of soln you are proposing earlier as well, but I have used variadic because otherwise I would have had to make changes in `newMachine()` call everywhere with second parameter as nill, while using variadic doesn't require that.
And I thought it was better for `newMachine()` be the only helper function to create new machines , and not make a new helper function just for making new machine with annotations.

pkg/aws/core.go
```diff
@@ -169,6 +172,13 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		}
 	}
 
+	placement, err := getPlacementObj(req)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if placement != nil {
+		inputConfig.Placement = placement
+	}
```
prashanth26, line 181:
```suggestion
	} else if placement != nil {
		inputConfig.Placement = placement
	}
```

pkg/aws/core_test.go
```diff
@@ -302,10 +306,26 @@ var _ = Describe("MachineServer", func() {
...
+					machineRequest: &driver.CreateMachineRequest{
+						Machine:      newMachine(0, annotations),
+						MachineClass: newMachineClass(providerSpec),
+						Secret:       providerSecret,
+					},
+				},
+				expect: expect{
+					errToHaveOccurred: false,
+					machineResponse: &driver.CreateMachineResponse{
+						ProviderID: "aws:///eu-west-1/i-0123456789-0",
```
prashanth26, line 320:
```suggestion
						ProviderID: "aws:///eu-west-1/i-0123456789-0/placement={affinity:host,availabilityZone:eu-west-1a,tenancy:host}"
```
You can update the mocking function to return something like this for validating this test case.

himanshu-kun:
I have made the changes as requested. one extra thing I did was updating the expected error message of test case `Machine creation request for capacity reservations fails if more than one type given` as it didn't match with the error message returned from validation.
Now all the testcases are running properly
I have also tested it on my local cluster with `m5.large` machine because the `canary cluster` has been taken down and I can no longer use that dedicated host.

prashanth26 (approved):
Thanks for these tests and changes.
/lgtm

kubernetes/machine-deployment.yaml
```diff
@@ -31,4 +31,8 @@ spec:
       class:
         kind: MachineClass # Machine class template used to create machine, could be AWS/GCP/Azure/Other-cloud-providers
         name: test-mc # Name of the machine class
-        namespace: default
\ No newline at end of file
+        namespace: default
+    # nodeTemplate: 
+    #   metadata:
+    #     annotations:  
+    #       machine.sapcloud.io/awsPlacement: '{ "affinity": "host", "availabilityZone", "eu-west-1a", "tenancy": "host"}' # Optional: it allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled; affinity can be host only if tenancy is host
```
prashanth26, line 38:
```suggestion
    #       machine.sapcloud.io/awsPlacement: '{ "affinity": "host", "availabilityZone":"eu-west-1a", "tenancy": "host"}' # Optional: it allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled; affinity can be host only if tenancy is host
```

kubernetes/machine.yaml
```diff
@@ -14,3 +14,7 @@ spec:
     kind: MachineClass
     name: test-mc
     namespace: default
+# nodeTemplate: 
+#   metadata:
+#     annotations:  
+#       machine.sapcloud.io/awsPlacement: '{ "affinity": "host", "availabilityZone", "eu-west-1a", "tenancy": "host"}' # Optional: it allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled; affinity can be host only if tenancy is host
```
prashanth26, line 20:
```suggestion
#       machine.sapcloud.io/awsPlacement: '{ "affinity": "host", "availabilityZone":"eu-west-1a", "tenancy": "host"}' # Optional: it allows placing machine on a dedicated host in an availability zone,assuming dedicated host has auto-placement enabled; affinity can be host only if tenancy is host
```

prashanth26 (commented):
@himanshu-kun - One small docu error I missed earlier.

pkg/aws/core.go
```diff
@@ -197,6 +206,26 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+//returns Placement Object required in ec2.RunInstancesInput
+func getPlacementObj(req *driver.CreateMachineRequest) (*ec2.Placement, error) {
+	placementobj := &ec2.Placement{}
+
+	requestAnnotations := req.Machine.Spec.NodeTemplateSpec.ObjectMeta.Annotations
+
+	if placementAnnotation, ok := requestAnnotations[awsPlacement]; ok && placementAnnotation != "" {
+		placementAnnotationsRaw := []byte(placementAnnotation)
+		err := json.Unmarshal(placementAnnotationsRaw, placementobj)
+		if err != nil {
```
amshuman-kr, line 218:
```suggestion
		if err := json.Unmarshal(placementAnnotationsRaw, placementobj); err != nil {
```

pkg/aws/core.go
```diff
@@ -197,6 +206,26 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+			return nil, err
+		}
+	}
+
+	if *placementobj == (ec2.Placement{}) {
```
amshuman-kr, line 223: Nice check.

pkg/aws/core_test.go
```diff
@@ -50,6 +50,9 @@ var _ = Describe("MachineServer", func() {
 			"userData":                []byte("dummy-user-data"),
 		},
 	}
+	annotations := map[string]string{
+		awsPlacement: "{ \"affinity\": \"host\", \"availabilityZone\": \"eu-west-1a\", \"tenancy\": \"host\"}",
```
amshuman-kr, line 54:
```suggestion
		awsPlacement: `{ "affinity": "host", "availabilityZone": "eu-west-1a", "tenancy": "host"}`,
```

pkg/aws/core_test.go
```diff
@@ -347,6 +367,186 @@ var _ = Describe("MachineServer", func() {
...
+				} else {
+					Expect(err).ToNot(HaveOccurred())
+					Expect(obj).To(Equal(data.expect.placementobj))
+				}
+			},
+			Entry("when Availability zone is set and affinity and tenancy set to host", &data{
+				setup: setup{
+					objectmeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							awsPlacement: "{ \"affinity\": \"host\", \"availabilityZone\": \"eu-west-1a\", \"tenancy\": \"host\"}",
```
amshuman-kr, line 403:
```suggestion
							awsPlacement: `{ "affinity": "host", "availabilityZone": "eu-west-1a", "tenancy": "host"}`,
```

pkg/aws/core_test.go
```diff
@@ -347,6 +367,186 @@ var _ = Describe("MachineServer", func() {
...
+						Affinity:         getStringPtr(ec2.AffinityHost),
+						Tenancy:          getStringPtr(ec2.TenancyHost),
+					},
+				},
+			}),
+			Entry("when Availability zone is not set but affinity and tenancy set to host", &data{
+				setup: setup{
+					objectmeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							awsPlacement: "{ \"affinity\": \"host\", \"tenancy\": \"host\"}",
```
amshuman-kr, line 427:
```suggestion
							awsPlacement: `{ "affinity": "host", "tenancy": "host"}`,
```

pkg/aws/core_test.go
```diff
@@ -347,6 +367,186 @@ var _ = Describe("MachineServer", func() {
...
+				},
+				expect: expect{
+					errToHaveOccurred: false,
+				},
+			}),
+			Entry("when Availability zone is set and affinity and tenancy set to value default", &data{
+				setup: setup{
+					objectmeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							awsPlacement: "{ \"affinity\": \"default\", \"availabilityZone\": \"eu-west-1a\", \"tenancy\": \"default\"}",
```
amshuman-kr, line 504:
```suggestion
							awsPlacement: `{ "affinity": "default", "availabilityZone": "eu-west-1a", "tenancy": "default"}`,
```

himanshu-kun:
@prashanth26 i have made the changes , pls check

pkg/aws/core_test.go
```diff
@@ -347,6 +367,186 @@ var _ = Describe("MachineServer", func() {
...
+						AvailabilityZone: getStringPtr("eu-west-1a"),
+						Tenancy:          getStringPtr(ec2.TenancyDefault),
+					},
+				},
+			}),
+			Entry("when just hostId and partitionNumber is set", &data{
+				setup: setup{
+					objectmeta: v1.ObjectMeta{
+						Annotations: map[string]string{
+							awsPlacement: "{ \"hostId\": \"h-0123b456af7f89123\", \"partitionNumber\": 7}",
```
amshuman-kr, line 528:
```suggestion
							awsPlacement: `{ "hostId": "h-0123b456af7f89123", "partitionNumber": 7}`,
```

amshuman-kr (approved):
Thanks for the changes @himanshu-kun. The PR looks really good now!

Just small stylistic suggestions. Otherwise LGTM.

pkg/aws/core_test.go
himanshu-kun: yes ,this way looks much cleaner , thnx :)

amshuman-kr (approved):
LGTM

prashanth26 (approved):

PR #39, opened 2021-07-28
Disable SrcAndDestCheck on NAT Instances

### LINKED ISSUE #36: Make src/dst checks configurable on awsmachineclass
To support calico's "CrossSubnet" mode on gardener clusters for provider type `aws` the src and dst checks (https://docs.aws.amazon.com/vpc/latest/userguide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) have to be disabled on machine network interface level. Therefore I would like to expose a field in the `awsmachineclass` (https://github.com/gardener/machine-controller-manager/blob/master/pkg/apis/machine/v1alpha1/aws_machineclass_types.go#L179) which contains a boolean value which describes if the src/dst checks on the interface level of the machines are enabled or disabled.
I would suggest something like 
```
// Describes a network interface.
// Please also see https://docs.aws.amazon.com/goto/WebAPI/ec2-2016-11-15/MachineAWSNetworkInterfaceSpecification
type AWSNetworkInterfaceSpec struct { 
   // If set to false, source and destination checks are disabled, default is true
   SrcAndDstChecksEnabled: bool `json:"srcAndDstChecksEnabled,omitempty"`
}
```
WDYT?
/cc: @zanetworker

AxiomSamarth:
This PR is needed to support calico's "CrossSubnet" mode on gardener clusters for provider type `aws` the src and dst checks. This PR introduces a new boolean field in the providerSpec of the MachineClass for AWS called `SrcAndDstChecksEnabled` which when set to `false` disabled the `SrcAndDestCheck` on that AWS EC2 instance.

Fixes #36 

**Release note**:

```feature user
A new boolean field called `SrcAndDstChecksEnabled` in the providerSpec of the MachineClass for AWS is introduced. The default value of this flag is `true` which retains current behavior. However, on setting this flag to `false` these checks are disabled.  
```


pkg/aws/core.go
```diff
@@ -335,6 +343,20 @@ func (d *Driver) GetMachineStatus(ctx context.Context, req *driver.GetMachineSta
...
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		err = disableSrcAndDestCheck(svc, requiredInstance.InstanceId)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
+
```
kon-angelo, line 359: Isn't it a bit weird that the `getMachineStatus` call has a side-effect and updates the instance?

AxiomSamarth: Indeed, @kon-angelo. You are correct. 

In order to modify this attribute, the AWS instance needs to be `Running`. So, if the `MachineController` fails just after the machine creation on AWS and just before the modification of `SrcAndDstCheck` attribute, then there would be no way MCM when restarted will know that this is missed. 

So, even though `GetMachineStatus()` driver method is expected to work atomically only to get the machine status, we have introduced this where it actually updates the machine attribute. However, it is covered & mentioned in another story [here](https://github.com/gardener/machine-controller-manager/issues/635) where it will be handled better in the future.

kon-angelo (approved):
/lgtm

DockToFuture (approved):
/lgtm

pkg/aws/core_util.go
```diff
@@ -68,6 +69,24 @@ func decodeProviderSpecAndSecret(machineClass *v1alpha1.MachineClass, secret *co
...
+		SourceDestCheck: &ec2.AttributeBooleanValue{
+			Value: pointer.BoolPtr(false),
+		},
+	}
+
+	_, err := svc.ModifyInstanceAttribute(srcAndDstCheckEnabled)
+	if err != nil {
+		return err
+	}
+	klog.V(3).Infof("Successfully disabled Source/Destination check on NAT instance %s.", *instanceID)
```
prashanth26, line 86: Minor log fix. This is a normal VM feature. NAT is only a use-case of this feature.
```suggestion
	klog.V(3).Infof("Successfully disabled Source/Destination check on instance %s.", *instanceID)
```

prashanth26 (approved):
Minor NIT. 
/lgtm otherwise. 
/squash

prashanth26 (approved):

PR #41, opened 2021-08-31
init upgrade prs

jkrayl:
This pr will enable upgrade pull requests which update a VERSION file in the repo.


.ci/component_descriptor
```diff
@@ -0,0 +1,15 @@
...
+  export MAIN_REPO_DIR="$(readlink -f "$(dirname "${0}")/..")"
+fi
+MCM_VERSION_FILEPATH="${MAIN_REPO_DIR}/MCM_VERSION"
+
+export VERSION="$(cat "${MCM_VERSION_FILEPATH}")"
+
+${ADD_DEPENDENCIES_CMD} \
+  --component-dependencies "{\"name\":\"github.com/gardener/machine-controller-manager\",\"version\":\"$VERSION\"}"
+
+mv "${BASE_DEFINITION_PATH}" "${COMPONENT_DESCRIPTOR_PATH}"
```
jkrayl, line 15: This script creates a simple component descriptor which has the machine-controller-manager as dependency

.ci/pipeline_definitions
```diff
@@ -33,6 +33,11 @@ machine-controller-manager-provider-aws:
     pull-request:
       traits:
         pull-request: ~
+    create-upgrade-prs:
+      traits:
+        cronjob:
+          interval: '5m'
+        update_component_deps: ~
```
jkrayl, line 40: This code snippet will add a new pipeline job which will poll for new versions of dependent components. Since the  machine controller manager is the only dependency only its version will be checked

.ci/set_dependency_version
```diff
@@ -0,0 +1,31 @@
...
+
+MCM_FILEPATH="${MAIN_REPO_DIR}/MCM_VERSION"
+
+if [ ! -f "${MCM_FILEPATH}" ]; then
+  echo "error: no such file: {MCM_FILEPATH}"
+  exit 1
+fi
+
+echo -n "${DEPENDENCY_VERSION}" > "${MCM_FILEPATH}"
+echo "set dependency-version of ${DEPENDENCY_NAME} to ${DEPENDENCY_VERSION}"
```
jkrayl, line 31: Script which will be called by the upgrade_pr job in the pipeline. It checks whether the upgraded component is the mcm component and writes the new version in a file. You can extend the script with your own logic or call other scripts. The complete file system diff at the end of the script will be committed.
https://gardener.github.io/cc-utils/traits/update_component_deps.html#component-upgrade-contract

AxiomSamarth: Thank you so much for this PR @jkrayl Loved this! 

I would also want the automated PR to include the vendored files from that new version of MCM. I have extended the bash script as you suggested to the achieve the same. Could you review it once and if you are convinced that the new set of commands would work then please add the suggested commit to this PR?

```suggestion
echo "set dependency-version of ${DEPENDENCY_NAME} to ${DEPENDENCY_VERSION}"

cd ${MAIN_REPO_DIR}
old_version=$(cat go.mod | grep github.com/gardener/machine-controller-manager | xargs)
new_version="github.com/gardener/machine-controller-manager ${DEPENDENCY_VERSION}"
sed -i -- 's#'"${old_version}"'#'"${new_version}"'#g' go.mod

make revendor
```

AxiomSamarth: @jkrayl Any view or comments on this?

jkrayl: lgtm :) 
I haven't tested it but the changed files will be committed at the end of the job so this will work

MCM_VERSION
```diff
@@ -0,0 +1 @@
+v0.40.0
```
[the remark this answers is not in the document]
AxiomSamarth:
```suggestion
v0.39.0
```

.ci/set_dependency_version
AxiomSamarth: Thank you, Johannes for the response. I will test it out with the idea you suggested.

AxiomSamarth (approved):
/lgtm

AxiomSamarth:
@prashanth26 @himanshu-kun 

I am merging this PR to test the new pipeline to auto vendor the latest MCM releases.

####  How is this being tested?
We have reduced the version of the version file to `0.39.0` so the job thinks we have an older version of MCM in this MCM-AWS. When we merge the PR, concourse will render the pipeline and add a new job that compares your local version to the released version and should see that there is a higher version and create a PR. Then we will see whether this script works or not.

prashanth26:
> @prashanth26 @himanshu-kun
> 
> I am merging this PR to test the new pipeline to auto vendor the latest MCM releases.
> 
> #### How is this being tested?
> We have reduced the version of the version file to `0.39.0` so the job thinks we have an older version of MCM in this MCM-AWS. When we merge the PR, concourse will render the pipeline and add a new job that compares your local version to the released version and should see that there is a higher version and create a PR. Then we will see whether this script works or not.

Guessing it didn't work?

PR #48, opened 2021-09-14
Updates test step of pipeline to use testmachinery base-step image

himanshu-kun:
tools like /cc/utils/cli.py will now be available in the container running the test step. `cli.py` can be used to access secret server.
Fixes #

**Release note**:

```other operator
NONE
```


AxiomSamarth (approved):
/lgtm

PR #52, opened 2021-09-15
Dummy Competitor PR

himanshu-kun:
This is a PR which is used to simulate the scenario where one PR is already running its IT on the clusters and a new PR also is raised . The IT for the latter should wait until the IT of the former are done...waiting will be atmost for 60min.
Fixes #

**Release note**:

```other operator
NONE
```


AxiomSamarth:
Closing this one as I believe the purpose of this PR is already fulfilled.

PR #58, opened 2021-12-22
Handle Eventual Consistency Problem in Create Driver Call

### LINKED ISSUE #54: Handle the `Eventual Consistency' during Machine creation
**What happened**:
MCM triggers creation of multiple instances on AWS for the same backing machine object due to AWS' 'Eventual Consistency` implementation. 

**What you expected to happen**:
MCM should not create multiple instances for the same backing machine object. 

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know**:
This needs to be probably handled in `CreateMachine` method by introducing a `DescribeInstance()` method with an exponential backoff mechanism with a timeout of 2 minutes as suggested by AWS docs. This needs to be checked after creation of instance and before returning the response from `CreateMachine` method.

**Environment**:

himanshu-kun:
This PR takes care of problem where due to eventual consistency problem of AWS multiple instances get created
Fixes #54 

I have made `maxElapsedTimeInBackoff` as a variable and not a `const` because I change this for the unit tests so that for unit-tests the retrying interval is 10sec and not 5min.
**Release note**:

```bugfix user
A problem is resolved where multiple VMs were created by the driver in case the APIs on the AWS are not consistent for some time.
```

### DIFF
--- a/pkg/aws/core_util.go
+++ b/pkg/aws/core_util.go
@@ -159,6 +159,19 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
 	return instances, nil
 }
 
+func confirmInstanceById(svc ec2iface.EC2API, instanceID string) (bool,error){
+	input := ec2.DescribeInstancesInput{
+		InstanceIds: []*string{&instanceID},
+	}
+
+	_, err := svc.DescribeInstances(&input)
+	if err!=nil{
+		return false,err
+	}
+
+	return true,nil
+}
+
 func (d *Driver) generateBlockDevices(blockDevices []api.AWSBlockDeviceMappingSpec, rootDeviceName *string) ([]*ec2.BlockDeviceMapping, error) {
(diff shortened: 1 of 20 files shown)


AxiomSamarth (commented):
/lgtm

pkg/aws/core_util.go
```diff
@@ -159,6 +159,19 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
 	return instances, nil
 }
 
+func confirmInstanceById(svc ec2iface.EC2API, instanceID string) (bool,error){
```
AxiomSamarth, line 162:
```suggestion
func confirmInstanceByID(svc ec2iface.EC2API, instanceID string) (bool,error){
```

AxiomSamarth (changes requested):

kon-angelo (approved):
/lgtm

dkistner (approved):
/lgtm

PR #59, opened 2021-12-23
Machine object deletion will proceed if backing VM not found

### LINKED ISSUE #56: Failing to delete machine which instance id is not found in the infrastruture
**What happened**:
A shoot cluster has been set in deletion, but it fails with
```
Flow "Shoot cluster deletion" encountered task errors: [task "Waiting until shoot worker nodes have been terminated" failed: Failed to delete Worker shoot--foo--bar/bar: Error deleting worker: Failed while waiting for all machine resources to be deleted: 'machine shoot--foo--bar-aws-r5d-large-z1-57655-wzp5n failed: VM deletion failed due to - machine codes error: code = [Internal] message = [InvalidInstanceID.NotFound: The instance ID 'i-0c2f23c5d881ca0f6' does not exist
	status code: 400, request id: 553fa61f-35a5-4fd2-ad1e-15d30dd2cc10]. Aborting operation. Initiate VM deletion']
```

There is indeed no such instance in AWS
``` bash
$ aws ec2 describe-instances --instance-ids i-0c2f23c5d881ca0f6 --output json
{
    "Reservations": []
}
```

The error originates from the respective machine
```bash
$ k get machine shoot--foo--bar-aws-r5d-large-z1-57655-wzp5n -o json | jq  .status.lastOperation.description -cr
VM deletion failed due to - machine codes error: code = [Internal] message = [InvalidInstanceID.NotFound: The instance ID 'i-0c2f23c5d881ca0f6' does not exist
        status code: 400, request id: c5e98927-903b-4f61-83d8-b2c852c75c68]. Aborting operation. Initiate VM deletion
```

From this point of view, it looks like this VM does not exist anymore in the infrastructure.

**What you expected to happen**:
The machine object to be deleted when the respective VM no longer exist on the infrastructure.

**How to reproduce it (as minimally and precisely as possible)**:
N/A
**Anything else we need to know**:

**Environment**:
AWS

himanshu-kun:
Till now for cases where a machine obj has ProviderID but the instance on provider is not found , deletion had been be retried.
With This PR if the AWS api returns `InvalidInstanceIDNotFound` error (which was earlier taken as an Internal error and not handled) , the instance deletion is assumed to be success.

Eventual consistency can't occur in this case as [another PR](https://github.com/gardener/machine-controller-manager-provider-aws/pull/58) handles that case for CreateMachine and so it can't reach DeleteMachine.
Also in worst case if instance still exists and we assumed its deletion with this new logic, orphan VM collection will deal with it.

Fixes #56 

I have changed the mockClient functions to return `InvalidInstanceIDNotFound` error in case instance is not found (which was not done earlier) . This mimics the actual behavior of AWS API.
The existing testcases are changed accordingly
Also wait for 5min is NOT introduced, as that would deviate the code from other providers a lot.

**Release note**:

```bugfix user
The machine obj will be deleted if the AWS API indicate absence of backing instance. Earlier retrying used to happen, which led to cases where machine obj never got deleted.
```

### DIFF
--- a/pkg/mockclient/mockclient.go
+++ b/pkg/mockclient/mockclient.go
@@ -163,7 +163,7 @@ func (ms *MockEC2Client) DescribeInstances(input *ec2.DescribeInstancesInput) (*
 			}
 		}
 		if !found {
-			return nil, fmt.Errorf("Couldn't find any instance matching requirement")
+			return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs do not exist")
 		}
 	} else {
 
@@ -191,12 +191,6 @@ func (ms *MockEC2Client) TerminateInstances(input *ec2.TerminateInstancesInput) 
 			ec2.UnsuccessfulInstanceCreditSpecificationErrorCodeInvalidInstanceIdMalformed, "",
 			fmt.Errorf("Termination of instance errorred out"),
 		)
-	} else if *input.InstanceIds[0] == InstanceDoesntExistError {
-		// If instance with instance ID doesn't exist
-		return nil, awserr.New(
-			ec2.UnsuccessfulInstanceCreditSpecificationErrorCodeInvalidInstanceIdNotFound, "",
-			fmt.Errorf("Instance with instance-ID doesn't exist"),
-		)
 	}
 
 	var desiredInstance ec2.Instance
@@ -217,7 +211,7 @@ func (ms *MockEC2Client) TerminateInstances(input *ec2.TerminateInstancesInput) 
 	ms.FakeInstances = &newInstanceList
 
 	if !found {
-		return nil, fmt.Errorf("Couldn't find instance with given instance-ID %s", *input.InstanceIds[0])
+		return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs do not exist")
 	}
 
 	return &ec2.TerminateInstancesOutput{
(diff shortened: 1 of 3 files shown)


himanshu-kun:
@AxiomSamarth I would like to merge this PR before PR #58 , as both have clashing changes. So rebase will be req.

pkg/mockclient/mockclient.go
```diff
@@ -163,7 +163,7 @@ func (ms *MockEC2Client) DescribeInstances(input *ec2.DescribeInstancesInput) (*
 			}
 		}
 		if !found {
-			return nil, fmt.Errorf("Couldn't find any instance matching requirement")
+			return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs donot exist")
```
AxiomSamarth, line 166: Nit:
```suggestion
			return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs do not exist")
```

pkg/mockclient/mockclient.go
```diff
@@ -217,7 +211,7 @@ func (ms *MockEC2Client) TerminateInstances(input *ec2.TerminateInstancesInput)
 	ms.FakeInstances = &newInstanceList
 
 	if !found {
-		return nil, fmt.Errorf("Couldn't find instance with given instance-ID %s", *input.InstanceIds[0])
+		return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs donot exist")
```
AxiomSamarth, line 214:
```suggestion
		return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs do not exist")
```

pkg/mockclient/mockclient.go
```diff
@@ -270,7 +257,7 @@ func (ms *MockEC2Client) StopInstances(input *ec2.StopInstancesInput) (*ec2.Stop
 	}
 
 	if !found {
-		return nil, fmt.Errorf("Couldn't find any instance matching requirement")
+		return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs donot exist")
```
AxiomSamarth, line 260:
```suggestion
		return nil, fmt.Errorf("InvalidInstanceID.NotFound: The instance IDs do not exist")
```

AxiomSamarth (commented):
/lgtm there is just one quick nit

himanshu-kun:
The `test` step didn't pass because there has been a major version jump for `ginkgo` and so tests don't work with that. A fixing PR to use `gingko v1.16.5` has been raised to deal with it . See [this](https://github.com/gardener/machine-controller-manager-provider-aws/pull/62)

I verified that the integration tests and unit tests pass locally and so have merged the PR as [this](https://github.com/gardener/machine-controller-manager-provider-aws/pull/62) PR needed some discussion before merging

PR #61, opened 2021-12-29
set instance metadata HttpPutResponseHopLimit to 2

tareqhs:

Set instance metadata http response hop limit to 2. This allows using EC2 IAM Roles from inside pods running on aws instances.

More info here: https://github.com/gardener/gardener-extension-provider-aws/issues/468
This is approach is taken by AWS's own EKS.
https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/
> Now, newly launched and any updated EKS managed node groups will be configured with a metadata token response hop limit set to 2.

**Release note**:
```feature user
Instance metadata http PUT response hop limit is set to 2. Allows using EC2 IAM roles.
```

### DIFF
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -45,6 +45,9 @@ const (
 	// awsEBSDriverName is the name of the CSI driver for EBS
 	awsEBSDriverName = "ebs.csi.aws.com"
 	awsPlacement     = "machine.sapcloud.io/awsPlacement"
+	// InstanceMetadataResponseHopLimit HTTP PUT response hop limit for instance metadata requests
+	// the default is 1 which makes it inaccessible from inside containers
+	InstanceMetadataResponseHopLimit = 2
 )
 
 // NewAWSDriver returns an empty AWSDriver object
@@ -154,6 +157,9 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		IamInstanceProfile:  iam,
 		NetworkInterfaces:   networkInterfaceSpecs,
 		TagSpecifications:   []*ec2.TagSpecification{tagInstance, tagVolume},
+		MetadataOptions: &ec2.InstanceMetadataOptionsRequest{
+			HttpPutResponseHopLimit: aws.Int64(InstanceMetadataResponseHopLimit),
+		},
 	}
 
 	// Set the AWS Capacity Reservation target. Using an 'open' preference means that if the reservation is not found, then


himanshu-kun (commented):
lgtm

Would like to hold its merge until an approval is given for the soln. mentioned in [this](https://github.com/gardener/gardener-extension-provider-aws/issues/468) issue.

himanshu-kun:
need/second-opinion @AxiomSamarth

AxiomSamarth:
Hey @tareqhs,

Thanks for the contribution. I have a few doubts or concerns & I would appreciate if you could help me understand more.

1. By changing the `CreateMachine` driver method here, the IMDSv2 will be adopted/offered to all the customers using Gardener. We would like to know if this configuration is something only needed by your team at the moment. Because, if I understand it correctly from the docs, then adopting IMDSv2 may also require some efforts to transition from IMDSv1 to IMDSv2 (let's say, if customers have some other set of automation suite running against their Shoot. My reference doc [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#instance-metadata-transition-to-version-2)). The doc also said, `When you specify that IMDSv2 must be used, IMDSv1 no longer works`. 

2. Changing the meta configuration in `CreateMachine` driver method will only result in new instances to be configured with IMDSv2 and the old instances of the Shoot would continue to exist with IMDSv1. We would like to understand if such a `nodeGroup` setup be okay? Or is there a need to migrate even other instances in the `nodeGroup` to IMDSv2. 

I would also like to /invite @dkistner @kon-angelo here to this thread to seek an opinion if Extension AWS would be the right place to introduce this field so that it is handled in as an optionally configurable attribute.

tareqhs:
hi @AxiomSamarth 
1. Setting option `HttpToken` controls whether IMDSv2 is forced (tokens required) or both versions work. My proposal does not modify this option. I have already tested the endpoints in existing gardener machines without sending tokens and they work fine.
1. I think we should modify existing instances too. This can be done in-place without require replacing the instances via the `ModifyInstanceMetadataOptions` API.

It seems nice to allow forcing IMDSv2 via customizing options, talking about my other proposal in the original issue.

AxiomSamarth:
Hi @tareqhs,

>My proposal does not modify this option.

Okay, my bad. So, here the IMDS version is untouched, of course. Only the PUT response hop limit is updated to 2 which default today is 1.

tareqhs:
@AxiomSamarth correct.

AxiomSamarth:
@tareqhs We had an internal discussion about this issue/PR. We request you to drive this issue via the extension provider and adapt the changes in the `CreateMachine` method referring the hop limit from the MachineClass instead of hardcoding it. 

Please let us know if you need any further assistance during the implementation.

himanshu-kun:
/close
closing the PR as the corresponding change needs to be made on extension provider https://github.com/gardener/gardener-extension-provider-aws

PR #62, opened 2022-01-03
Use ginkgo v1 for tests

himanshu-kun:
Recently there has been a major version jump in ginkgo, due to which tests are failing. 
This PR will enable use of latest `v1.16.5` of ginkgo for tests.

Fixes #

`go install` is used as it doesn't affect go.mod and `go get` doesn't support getting a particular version without affecting go.mod.
**Release note**:

```doc developer
Ginkgo version 1.16.5 is used for tests until changes compatible with v2.0.0 are made.
```

### DIFF
--- a/.ci/test
+++ b/.ci/test
@@ -16,6 +16,9 @@
 
 set -e
 
+# to mute ginkgo deprecation warnings
+export ACK_GINKGO_DEPRECATIONS=2.0.0
+
 # For the test step concourse will set the following environment variables:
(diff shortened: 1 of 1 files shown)


AxiomSamarth (approved):

himanshu-kun:
@AxiomSamarth  we might have to raise similar PRs on all providers. I wonder if there is anything like this done in other Gardener repos?

PR #68, opened 2022-03-07
SrcDestCheck changes done after confirming consistency

himanshu-kun:
Fix : earlier srcDestCheck related changes on instance were made right after instance creation.
Fixes #

Its a bug fix and will be hotfixed
**Release note**:

```bugfix user
srcDestCheck enable/disable for a instance is done after confirming API consistency
```


unmarshall (commented):
/lgtm - ideally there should have been a way for disabling NAT source/dest check while creating an instance and then it should have been handled by AWS instead of asking the consumer to first create an instance, wait for an instance to be up and running and then set this to false.

PR #89, opened 2022-06-16
Removing IOPS validation

rishabh-11:
This PR will allow setting IOPS for any volume type 
Fixes #

Testing on the dev landscape is completed
**Release note**:

```enhancement user
Users can now set IOPS for a GP3 volume type. There is no change for IO1. Validation of IOPS (i.e. whether it is allowed and is in the specified range for a volume type) is done on the AWS side, so feedback will arrive once the volume starts getting created.
```

### DIFF
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -160,7 +160,7 @@ type AWSEbsBlockDeviceSpec struct {
 	// and bursting, see Amazon EBS Volume Types (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html)
 	// in the Amazon Elastic Compute Cloud User Guide.
 	//
-	// Constraint: Constraint: IOPS should be a positive value.
+	// Constraint: IOPS should be a positive value.
 	// Validation of IOPS (i.e. whether it is allowed and is in the specified range for a particular volume type) is done on aws side.
 	//
 	// Condition: This parameter is required for requests to create io1 volumes;
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -153,18 +153,18 @@ type AWSEbsBlockDeviceSpec struct {
 	Encrypted bool `json:"encrypted,omitempty"`
 
 	// The number of I/O operations per second (IOPS) that the volume supports.
-	// For io1, this represents the number of IOPS that are provisioned for the
+	// For io1 and gp3, this represents the number of IOPS that are provisioned for the
 	// volume. For gp2, this represents the baseline performance of the volume and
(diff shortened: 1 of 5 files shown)


pkg/aws/apis/validation/validation.go
```diff
@@ -123,10 +123,9 @@ func validateBlockDevices(blockDevices []awsapi.AWSBlockDeviceMappingSpec, fldPa
 			allErrs = append(allErrs, field.Required(idxPath.Child("ebs.volumeSize"), "Please mention a valid EBS volume size"))
 		}
 
-		if disk.Ebs.VolumeType == awsapi.VolumeTypeIO1 && disk.Ebs.Iops <= 0 {
+		if disk.Ebs.Iops < 0 || (disk.Ebs.VolumeType == awsapi.VolumeTypeIO1 && disk.Ebs.Iops == 0) {
```
himanshu-kun, line 126:
```suggestion
		if disk.Ebs.Iops <= 0 || (disk.Ebs.VolumeType == awsapi.VolumeTypeIO1 && disk.Ebs.Iops == 0) {
```

pkg/aws/apis/aws_provider_spec.go
```diff
@@ -153,18 +153,18 @@ type AWSEbsBlockDeviceSpec struct {
...
+	// For io1 and gp3, this represents the number of IOPS that are provisioned for the
 	// volume. For gp2, this represents the baseline performance of the volume and
 	// the rate at which the volume accumulates I/O credits for bursting. For more
 	// information about General Purpose SSD baseline performance, I/O credits,
 	// and bursting, see Amazon EBS Volume Types (http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html)
 	// in the Amazon Elastic Compute Cloud User Guide.
 	//
-	// Constraint: Range is 100-20000 IOPS for io1 volumes and 100-10000 IOPS for
-	// gp2 volumes.
+	// Constraint: Constraint: IOPS should be a positive value.
```
himanshu-kun, line 163:
```suggestion
	// Constraint: IOPS should be a positive value.
```

himanshu-kun (changes requested):

pkg/aws/apis/validation/validation.go
rishabh-11: `disk.Ebs.Iops = 0` means that the IOPS field was not set in the configuration. This is invalid only in the case of IO1 volumes and not for GP3.

himanshu-kun: hmm. right, then pls ignore this suggestion

dkistner:
looks good to me. Once the pending test on dev are done I guess we are good to move on.

himanshu-kun (approved):
/lgtm

PR #90, opened 2022-07-14
Change base image from alpine to distroless

ialidzhikov:
/area security
/kind enhancement

Similar to https://github.com/gardener/machine-controller-manager-provider-openstack/pull/62

**Release note**:

```other operator
machine-controller-manager-provider-aws now uses `distroless` instead of `alpine` as a base image.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-#############      builder                                  #############
+#############      builder                          #############
 FROM golang:1.17.9 AS builder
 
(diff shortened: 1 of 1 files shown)


dkistner:
Looks good to me.
I guess the test is failing as the credentials of the test cluster are rotated. Can you check @himanshu-kun?

himanshu-kun (approved):
/lgtm

PR #102, opened 2023-01-24
Vendor `MCM v0.48.0`, Update golang version to 1.19.5

rishabh-11:
Vendor `MCM v0.48.0`. Update golang version to 1.19.5

Fixes #

**Release note**:

```other dependency
The following dependency is updated:
- github.com/gardener/machine-controller-manager v0.47.0 -> 0.48.0
```
```other operator
update golang version to 1.19.5
```


pkg/aws/aws_suite_test.go
```diff
@@ -53,14 +55,11 @@ func newMachine(
...
 
-	// Don't initialize providerID and node if setMachineIndex == -1
+	// Don't initialize providerID if setMachineIndex == -1
 	if setMachineIndex != -1 {
 		machine.Spec = v1alpha1.MachineSpec{
 			ProviderID: fmt.Sprintf("aws:///eu-west-1/i-0123456789-%d", setMachineIndex),
 		}
-		machine.Status = v1alpha1.MachineStatus{
-			Node: fmt.Sprintf("ip-%d", setMachineIndex),
-		}
```
himanshu-kun, line 63: we should replace this with `setting the node label`

himanshu-kun (changes requested):
a minor change requested. rest looks good

pkg/aws/aws_suite_test.go
rishabh-11: done

elankath:
Shouldn't this be something like:
```go
      machine.Labels = map[string]string{
                 v1alpha1.NodeLabelKey: fmt.Sprintf("ip-%d", setMachineIndex),
      }
```

rishabh-11:
> Shouldn't this be something like:
> 
> ```go
>       machine.Labels = map[string]string{
>                  v1alpha1.NodeLabelKey: fmt.Sprintf("ip-%d", setMachineIndex),
>       }
> ```

yes it should be. i have changed it now

himanshu-kun (approved):
/lgtm

PR #104, opened 2023-02-15
Drop `spec.KeyName` validation

AleksandarSavchev:
This PR drops validation for `KeyName` in `AWSProviderSpec`. Since [this PR] (https://github.com/gardener/gardener/pull/7188) it is possible to disable `sshAccess` for the worker nodes and by doing so, stop the `keyPair` rotation. In that case the `KeyName` would be an empty string.

Part of [gardener/gardener#7481](https://github.com/gardener/gardener/issues/7481)

This PR is required to be merged and released for the changes in [this draft.](https://github.com/gardener/gardener-extension-provider-aws/pull/704)

**Release note**:

```other operator
Dropped validation for `KeyName` in `AWSProviderSpec`.
```
```breaking user
`KeyName` in `AWSProviderSpec` struct has been changed to a `pointer`.
```

### DIFF
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -160,10 +160,13 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		MinCount:            aws.Int64(1),
 		MaxCount:            aws.Int64(1),
 		UserData:            &UserDataEnc,
-		KeyName:             aws.String(providerSpec.KeyName),
 		IamInstanceProfile:  iam,
 		NetworkInterfaces:   networkInterfaceSpecs,
 		TagSpecifications:   []*ec2.TagSpecification{tagInstance, tagVolume, tagNetworkInterface},
+	}
+
+	if len(providerSpec.KeyName) > 0 {
+		inputConfig.KeyName = aws.String(providerSpec.KeyName)
 	}
 
 	// Set the AWS Capacity Reservation target. Using an 'open' preference means that if the reservation is not found, then
@@ -240,7 +243,7 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 	return response, nil
 }
 
-//returns Placement Object required in ec2.RunInstancesInput
+// returns Placement Object required in ec2.RunInstancesInput
 func getPlacementObj(req *driver.CreateMachineRequest) (*ec2.Placement, error) {
 	placementobj := &ec2.Placement{}
 
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -165,8 +165,8 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		TagSpecifications:   []*ec2.TagSpecification{tagInstance, tagVolume, tagNetworkInterface},
 	}
 
-	if len(providerSpec.KeyName) > 0 {
-		inputConfig.KeyName = aws.String(providerSpec.KeyName)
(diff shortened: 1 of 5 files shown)


himanshu-kun (approved):
/lgtm

himanshu-kun:
Just curious, if ssh is disabled for the worker , then would there be another way to login to the machine?

pkg/aws/core.go
```diff
@@ -160,12 +160,15 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		MinCount:            aws.Int64(1),
 		MaxCount:            aws.Int64(1),
 		UserData:            &UserDataEnc,
-		KeyName:             aws.String(providerSpec.KeyName),
 		IamInstanceProfile:  iam,
 		NetworkInterfaces:   networkInterfaceSpecs,
 		TagSpecifications:   []*ec2.TagSpecification{tagInstance, tagVolume, tagNetworkInterface},
 	}
 
+	if len(providerSpec.KeyName) > 0 {
```
unmarshall, line 168: Since `KeyName` is not optional, shouldn't the `AwsProviderSpec` be update to reflect that?
```go
type AWSProviderSpec struct {
   ...
   // +optional
   KeyName *string
}
```
Will this cause an issue for any consumer?

AleksandarSavchev:
@himanshu-kun the machines would still be accessible by pod-host network. Tools like [node-shell](https://github.com/kvaps/kubectl-node-shell) would still work.

pkg/aws/core.go
AleksandarSavchev: The `// +optional` is not needed since it is not used for the other optional fields in the struct `CapacityReservationTarget` and `SpotPrice`. Changing `KeyName` to be a pointer would be reasonable, however that would be a breaking change for whoever has imported this struct and I am not familiar with the consumers.

unmarshall: We checked that `AwsProviderSpec` is not used outside this repo. So changing it to pointer to string should be fine.
@himanshu-kun would you also agree?

himanshu-kun: Yes I agree, we should do that. Please adapt the release note to note the field name change as `breaking change`

tedteng:
> @himanshu-kun the machines would still be accessible by pod-host network. Tools like [node-shell](https://github.com/kvaps/kubectl-node-shell) would still work.

I understand node-shell still be accessible to the cluster node. as the node is Ready and join the cluster.

one thing I want to know,  if an issue VM not join the cluster(node not ready), are we still have a chance to use ssh to access the issue VM to find the VM log for troubleshooting purposes? then sshd disabled after the issue VM is fixed and joined the cluster automatically. Is that correct for my understanding?

unmarshall (approved):

AleksandarSavchev:
> I understand node-shell still be accessible to the cluster node. as the node is Ready and join the cluster.
> 
> one thing I want to know, if an issue VM not join the cluster(node not ready), are we still have a chance to use ssh to access the issue VM to find the VM log for troubleshooting purposes? then sshd disabled after the issue VM is fixed and joined the cluster automatically. Is that correct for my understanding?

With `sshAccess` disabled the creation of bastion resources is forbidden and the `sshd` service on these VMs is disabled/ not running. If you want to have ssh access to these node you can change the shoot definition to enabled `sshAccess`.

tedteng:
Thanks. yes, just realize the shoot YAML already changed(sshAccess enable) it doesn't matter if a new VM joins or not

PR #105, opened 2023-02-22
add instance metadatav2 support

kon-angelo:

Fixes MCM part of #https://github.com/gardener/gardener-extension-provider-aws/issues/468

**Release note**:

```feature operator
added support to modify instance metadata parameters for a VM through machineClass only during creation. This could be used to enable instance metadatav2 to containerised environment
```


pkg/aws/apis/validation/validation.go
```diff
@@ -191,6 +193,29 @@ func validateNetworkInterfaces(networkInterfaces []awsapi.AWSNetworkInterfaceSpe
...
 
+func validateInstanceMetadata(metadata *awsapi.InstanceMetadataOptions, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if metadata == nil {
+		return allErrs
+	}
+
+	if metadata.HttpPutResponseHopLimit != nil {
+		if *metadata.HttpPutResponseHopLimit < 0 || *metadata.HttpPutResponseHopLimit > 64 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpPutResponseHopLimit"), *metadata.HttpPutResponseHopLimit, "Only values between 0 and 64 are accepted"))
```
himanshu-kun, line 204:
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpPutResponseHopLimit"), *metadata.HttpPutResponseHopLimit, "Only values between 0 and 64, both included, are accepted"))
```

pkg/aws/apis/validation/validation.go
```diff
@@ -191,6 +193,29 @@ func validateNetworkInterfaces(networkInterfaces []awsapi.AWSNetworkInterfaceSpe
...
+		}
+
+		if metadata.HttpEndpoint != nil && *metadata.HttpEndpoint != "disabled" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpEndpoint"), *metadata.HttpEndpoint, "Only 'disabled' is valid value"))
```
himanshu-kun, line 208: technically `enabled` could be provided as well, so we should accept that value also.

pkg/aws/apis/validation/validation.go
```diff
@@ -191,6 +193,29 @@ func validateNetworkInterfaces(networkInterfaces []awsapi.AWSNetworkInterfaceSpe
...
+		}
+
+		if metadata.HttpTokens != nil && *metadata.HttpTokens != "required" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpTokens"), *metadata.HttpTokens, "Only 'required' is valid value"))
```
himanshu-kun, line 212: `optional` is also valid. Yes its default right now, so anyone would only set `required` but later `requried` could become default value.

pkg/aws/apis/aws_provider_spec.go
```diff
@@ -103,6 +103,8 @@ type AWSProviderSpec struct {
 
 	// Tags to be specified on the EC2 instances
 	Tags map[string]string `json:"tags,omitempty"`
+
+	InstanceMetadataOptions *InstanceMetadataOptions `json:"instanceMetadataOptions,omitempty"`
```
himanshu-kun, line 107: why do we need to define our own types and later copy values from here to the type provided by ec2? 
We have done this way earlier also but I don't get the reasoning. Is it to make sure that if a backward incompatible change happens to their struct then we don't have to make a lot of changes, like in extensions ?

himanshu-kun (changes requested):
minor comments

pkg/aws/apis/aws_provider_spec.go
kon-angelo: I think it makes sense to decouple the two specs. We may want to use custom json decoding, do some kind of translation in-between our spec and ec2 client, control the default behavior etc. This is also an "extreme" case where we want to expose all fields from the ec2 struct (which doesn't always make sense). 

In general I am in favor of having control over the API we expose even if inadvertently it will approximate the ec2 one.

himanshu-kun:
current PR doesn't address updaing the metadata options for already running/stopped (existing) VMs
Looking from an open-source point of view , all instances created by the machineClass should be updated once machineClass is updated, if the cloud allows for updating that VM parameter without restart. This is not currently supported overall, but `instance metadata` could be added once the support is enabled. Track https://github.com/gardener/machine-controller-manager/issues/635

pkg/aws/apis/aws_provider_spec.go
himanshu-kun: @unmarshall @rishabh-11 also had the same doubt. Do you guys agree with @kon-angelo's comment or still think that we should couple them.

pkg/aws/apis/validation/validation.go
kon-angelo: Added validation for the suggested values 👍

pkg/aws/apis/aws_provider_spec.go
unmarshall: When i look at the `AWSProviderSpec` i see other instances where another struct has been defined. Example `AWSNetworkInterfaceSpec` which holds the network interface spec fields which gardener configures. This is then used to create `ec2.InstanceNetworkInterfaceSpecification`. Similar is the case elsewhere as well. So what @kon-angelo has done is not out of the ordinary.

I think it is fine to declare our own structs. These structs will then represent fields that are today managed (set/configured) by extensions.

unmarshall, line 107: Can you please add a docstring to `InstanceMetadataOptions`?

pkg/aws/apis/validation/validation.go
```diff
@@ -188,6 +190,27 @@ func validateNetworkInterfaces(networkInterfaces []awsapi.AWSNetworkInterfaceSpe
...
+	if metadata.HTTPPutResponseHopLimit != nil && (*metadata.HTTPPutResponseHopLimit < 0 || *metadata.HTTPPutResponseHopLimit > 64) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("httpPutResponseHopLimit"), *metadata.HTTPPutResponseHopLimit, "Only values between 0 and 64, both included, are accepted"))
+	}
+
+	if metadata.HTTPEndpoint != nil {
+		allErrs = append(allErrs, validateStringValues(fldPath.Child("httpEndpoint"), *metadata.HTTPEndpoint, []string{awsapi.HTTPEndpointDisabled, awsapi.HTTPEndpointEnabled})...)
+	}
+
+	if metadata.HTTPTokens != nil {
+		allErrs = append(allErrs, validateStringValues(fldPath.Child("httpEndpoint"), *metadata.HTTPTokens, []string{awsapi.HTTPTokensRequired, awsapi.HTTPTokensOptional})...)
```
unmarshall, line 208:
```suggestion
		allErrs = append(allErrs, validateStringValues(fldPath.Child("httpTokens"), *metadata.HTTPTokens, []string{awsapi.HTTPTokensRequired, awsapi.HTTPTokensOptional})...)
```

pkg/aws/apis/validation/validation.go
```diff
@@ -219,3 +242,11 @@ func contains(arr []string, checkValue string) bool {
 	}
 	return false
 }
+
+func validateStringValues(fld *field.Path, s string, accepted []string) field.ErrorList {
```
unmarshall, line 246: I have seen this kind of function too many times. One could just use https://pkg.go.dev/golang.org/x/exp/slices which essentially contains a lot of such utilities and prevents duplication of code. For now you can continue using this.

unmarshall, line 246: This function could have just returned `Error` instead of `ErrorList` as there cannot be more than one error anyways.

pkg/aws/apis/validation/validation_test.go
```diff
@@ -1334,6 +1334,99 @@ var _ = Describe("Validation", func() {
 					},
 				},
 			}),
+			Entry("AWS machine class with instanceMetadata", &data{
+				setup: setup{},
+				action: action{
+					spec: &awsapi.AWSProviderSpec{
```
unmarshall, line 1340: It will make the tests more readable if you create a function which will create and initialize `AWSProviderSpec` and then in individual tests you can set `InstanceMetadataOptions` with different values.

pkg/aws/apis/validation/validation_test.go
```diff
@@ -1334,6 +1334,99 @@ var _ = Describe("Validation", func() {
 					},
 				},
 			}),
+			Entry("AWS machine class with instanceMetadata", &data{
```
unmarshall, line 1337:
```suggestion
			Entry("AWS machine class with instanceMetadata with valid instanceMetadata.httpPutResponseHopLimit", &data{
```

pkg/aws/apis/validation/validation_test.go
```diff
@@ -1334,6 +1334,99 @@ var _ = Describe("Validation", func() {
...
+				expect: expect{
+					errToHaveOccurred: true,
+					errList: field.ErrorList{
+						{
+							Type:     "FieldValueInvalid",
+							Field:    "providerSpec.instanceMetadata.httpPutResponseHopLimit",
+							BadValue: int64(72),
+							Detail:   "Only values between 0 and 64, both included, are accepted",
+						},
+					}},
```
unmarshall, line 1428: You have only written tests for hop limits, however validations have also been added for httpendpoint. Can you add those tests as well?

pkg/aws/apis/validation/validation.go
kon-angelo: True, but I think the `ErrorList` is much more convenient to consume as a type. You can simply do an `append` to merge your error. Otherwise your signature needs to return `*field.Error` and need additional `nil` checks for the consumer of the `ErrorList`.

pkg/aws/apis/validation/validation_test.go
kon-angelo: The whole file is basically written like that and I wanted to avoid having to refactor all the tests, but I can do that

kon-angelo: I changed the last tests I added as suggested

pkg/aws/apis/validation/validation.go
kon-angelo: I changed to use the `slices` package. I wasn't aware of it but we can use that instead.

pkg/aws/apis/validation/validation_test.go
kon-angelo: Added with latest commit

himanshu-kun (approved):
/lgtm
Earlier I was thinking make this enahancement https://github.com/gardener/gardener-extension-provider-aws/pull/709#discussion_r1122902301
But I guess keeping separate structs in separate repos would be more convenient?

unmarshall (approved):
lgtm

himanshu-kun:
> /lgtm Earlier I was thinking make this enahancement [gardener/gardener-extension-provider-aws#709 (comment)](https://github.com/gardener/gardener-extension-provider-aws/pull/709#discussion_r1122902301) But I guess keeping separate structs in separate repos would be more convenient?

Reasoning given here -> https://github.com/gardener/gardener-extension-provider-aws/pull/709#discussion_r1127517930

himanshu-kun (changes requested):
kindly also update the following file 
[kubernetes/machine-class.yaml](https://github.com/gardener/machine-controller-manager-provider-aws/blob/1b046c47b84e7164a444b655b639fcd63bb1dd6e/kubernetes/machine-class.yaml)

himanshu-kun (approved):
/lgtm
Thanks :)

PR #108, opened 2023-02-24
Add cve categorization for mcm-provider-aws

dkistner:
Add cve categorization for mcm-provider-aws.

The categorization was initially defined here and it transferred as defined to the mcm-provider-aws repository: 
- https://github.com/gardener/gardener-extension-provider-aws/pull/684

We move the cve categorization to the mcm-provider-aws repository as the categorization can't be taken into consideration when it is defined in the respective provider extension project. Once this behaviour has changed we can remove the categorisation here again.

**Release note**:
```other operator
CVE categorization for mcm-provider-aws has been added.
```


himanshu-kun (approved):
/lgtm

PR #128, opened 2023-07-18
Add `Ipv6AddressCount` and `Ipv6PrefixCount` to enable the assignment of an ipv6 address and an ipv6 prefix to instances.

axel7born:
This PR is needed to add IPv6 support to Gardener by obtaining globally routable IPv6 addresses from AWS. We use `Ipv6AddressCount` to automatically assign IPv6 addresses from the subnet range. Additionally, we use `Ipv6PrefixCount` to get delegated prefixes for the pod CIDR

Fixes #

**Release note**:

```feature operator
Add `Ipv6AddressCount` and `Ipv6PrefixCount` to enable the assignment of an ipv6 address and an ipv6 prefix to instances.
```


pkg/aws/core.go
```diff
@@ -243,6 +247,20 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
 
+	for i, netIf := range providerSpec.NetworkInterfaces {
+		networkInterfaceID := runResult.Instances[0].NetworkInterfaces[i].NetworkInterfaceId
+		if netIf.Ipv6PrefixCount != nil {
+			input := &ec2.AssignIpv6AddressesInput{
+				NetworkInterfaceId: networkInterfaceID,
+				Ipv6PrefixCount:    aws.Int64(*netIf.Ipv6PrefixCount),
+			}
+			_, err = svc.AssignIpv6Addresses(input)
+			if err != nil {
```
himanshu-kun, line 258: shouldn't we also check whether the networkInterface has IPv6 IPs assigned before getting ipv6 delegated prefixes for it ? or are you planning to have ipv6 delegated prefixes for nics with ipv4 addresses also?

pkg/mockclient/mockclient.go
```diff
@@ -122,6 +122,9 @@ func (ms *MockEC2Client) RunInstances(input *ec2.RunInstancesInput) (*ec2.Reserv
 	newInstance := ec2.Instance{
 		InstanceId:     &instanceID,
 		PrivateDnsName: &privateDNSName,
+		NetworkInterfaces: []*ec2.InstanceNetworkInterface{{
+			NetworkInterfaceId: aws.String("i-123456789"),
+		}},
```
himanshu-kun, line 127: why was this change needed?

himanshu-kun (changes requested):
I am not sure what combination of fields could be specified together , but guessing AWS shouldn't allow fields for ipv4 and ipv6 to be specified together. If thats the case kindly add validation to ensure both the 
`AssociatePublicIpAddress` -> for ipv4 and
the new fields `Ipv6AddressCount` (and `ipv6PrefixCount` if it holds) couldn't be specified together

pkg/aws/core.go
```diff
@@ -243,6 +247,18 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		}
 	}
 
+	for i, netIf := range providerSpec.NetworkInterfaces {
+		networkInterfaceID := runResult.Instances[0].NetworkInterfaces[i].NetworkInterfaceId
+		input := &ec2.AssignIpv6AddressesInput{
+			NetworkInterfaceId: networkInterfaceID,
```
ScheererJ: Do you want to inline `networkInterfaceID`?

```suggestion
		input := &ec2.AssignIpv6AddressesInput{
			NetworkInterfaceId: runResult.Instances[0].NetworkInterfaces[i].NetworkInterfaceId,
```

axel7born:
> I am not sure what combination of fields could be specified together , but guessing AWS shouldn't allow fields for ipv4 and ipv6 to be specified together. If thats the case kindly add validation to ensure both the `AssociatePublicIpAddress` -> for ipv4 and the new fields `Ipv6AddressCount` (and `ipv6PrefixCount` if it holds) couldn't be specified together

It shouldn't be a problem to specify a public IPv4 address together with an IPv6 address. In case of a bastion you still need a public IPv4 address, even if nodes have IPv6 addresses and private IPv4 addresses.
I just tested it by creating a bastion with both public IPv4 address and IPv6 address.

pkg/aws/core.go
axel7born: The `Ipv6PrefixCount` and `Ipv6AddressCount` can't be set in one request. However, having an Ipv6 address is not a prerequisite for getting a prefix.
We don't plan to have IPv6 prefixes for instances with only IPv4 addresses. Actually, I can't imagine how that would make sense. However, from AWS side it is still possible.
We wouldn't enable that combination in a `NodeClass`. 
Do we really need to validate that here?

pkg/mockclient/mockclient.go
axel7born: This change is needed as during unit tests we would crash here: https://github.com/gardener/machine-controller-manager-provider-aws/blob/ed01605d61036b3bd5e81917c1ee7998f5db5127/pkg/aws/core.go#L251
However, if we inline `networkInterfaceID` as suggested by @ScheererJ, we can remove it again.

pkg/aws/core.go
```diff
@@ -243,6 +247,19 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		}
 	}
 
+	for i, netIf := range providerSpec.NetworkInterfaces {
+		if netIf.Ipv6PrefixCount != nil {
+			input := &ec2.AssignIpv6AddressesInput{
+				NetworkInterfaceId: runResult.Instances[0].NetworkInterfaces[i].NetworkInterfaceId,
+				Ipv6PrefixCount:    aws.Int64(*netIf.Ipv6PrefixCount),
```
DockToFuture, line 254:
```suggestion
				Ipv6PrefixCount:    netIf.Ipv6PrefixCount,

```

pkg/aws/core.go
```diff
@@ -144,6 +144,10 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 			spec.DeleteOnTermination = aws.Bool(true)
 		}
 
+		if netIf.Ipv6AddressCount != nil {
+			spec.Ipv6AddressCount = aws.Int64(*netIf.Ipv6AddressCount)
```
DockToFuture, line 148:
```suggestion
						spec.Ipv6AddressCount = netIf.Ipv6AddressCount
```

pkg/aws/core.go
himanshu-kun: The concern is not from Gardener consumption side , but for users consuming mcm-aws directly. If they provide multiple nics specification in the providerSpec , like in order  { ipv4 , ipv6 with `ipv6prefixCount` set , ipv6 with `ipv6AddressCount` set, ipv4} and then in this for loop, `ipv6prefixCount` might get set up for an `ipv4` nic (depends on how `runResult.Instances[0].NetworkInterfaces[]` are arranged)... 

so we should deal with this either via validation or someway else.

axel7born: I was assuming that `DeviceIndex` would be the index in that array. However, I didn't find docs that explicitly state that this is really the case. Now, I explicitly look for a matching `DeviceIndex`.

himanshu-kun (commented):
The changes looks good now for `CreateMachine`, but as per the current design , mcm-providers cannot deal with a failure in a non-atomic create call. Since you are updating the network interfaces after the vm create call , any failure in them, would mean no calling back of `CreateMachine` in the coming reconciles (as the VM is already created)

Non-Atomic create has been done once before when `srcDestCheck` was made [configurable](https://github.com/gardener/machine-controller-manager-provider-aws/pull/68) and to tackle the above mentioned problem , update was also introduced in `GetMachineStatus` (which ofcourse is not according to the method's contract) to allow updates in upcoming reconciles.

We don't want to repeat the same in this case, but it depends on the urgency of this feature. So could you help us understanding if this feature has any hard deadlines?

axel7born:
From my point of view, this is not so urgent that we would need it in the next couple of weeks. There are other missing parts we haven't even started with. However, I'm not sure now how to move forward. Do you have an idea on how to solve the problem of non-atomic updates? Is that something we should do in this PR? On the other hand, if we merge this now, it won't be used in production very soon, but it would simplify further development.
WDYT @DockToFuture @ScheererJ

unmarshall:
If this is not urgent then i would recommend that we do not add more technical debt.  Due to https://github.com/gardener/machine-controller-manager-provider-aws/pull/39 we already have broken the contract of `GetMachineStatus`. The reasoning then was that it was urgent. Well this is only piling up now. The issue is that we do not yet have `UpdateMachine` in the Driver interface. This is exactly what you need here and that retains the atomicity of operations done via `CreateMachine` and also ensures that a `Get` does not do an update.

unmarshall:
@axel7born created https://github.com/gardener/machine-controller-manager/issues/840. We will discuss this in the planning today and assign it an appropriate priority. Once that is done then we will update on the timelines here.

elankath:
Hi @axel7born , we are adding a new `Driver.InitializeMachine` operation that can be implemented by providers to perform post-creation initialization of the VM. PR: https://github.com/gardener/machine-controller-manager-provider-aws/pull/128, Preview: https://github.com/gardener/machine-controller-manager/blob/e6b614b52016b3ff902d53eaa198bd1c9a436f16/docs/proposals/initialize-machine.md. 

Initialization stuff like disabling src/dest checks for NAT instances and assignment of Ipv6 addresses can be done here. The semantics of `Driver.GetMachineStatus` will also be changed to return a status error code if initialization of the instance is not yet done. Please let us know if this is suitable for this PR.  (will adjust your PR after our feature is merged)

axel7born:
Hi @elankath, thanks for the update.
For me this looks great!

elankath:
@axel7born sorry for long delay. We will be merging and releasing `Driver.InitMachine` shortly in one or two days (was delayed due to other things including vendor removal).  Should I take care of` Ipv6AddressCount` and `Ipv6PrefixCount` adoption within `InitMachine` myself or will you wish to take care of the adoption yourself ?

axel7born:
@elankath Thank's for the update. I would be happy if you could handle the adoption.

elankath:
@axel7born  could you summarize the testing that you had done and any additional tests that you would like to see. will carry out those tests and add a test log.

PR #132, opened 2023-10-16
Expose logs for eventual consistency checks

himanshu-kun:

Fixes #

**Release note**:

```bugfix operator
Logs for confirming eventual consistency are exposed. This helps to know if such check was successful or not.
```

### DIFF
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -223,7 +223,7 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
 		NodeName:   *runResult.Instances[0].PrivateDnsName,
 	}
 
-	klog.V(3).Infof("Waiting for VM with Provider-ID %q to be visible to all AWS endpoints", response.ProviderID)
+	klog.V(2).Infof("Waiting for VM with Provider-ID %q, for machine %q to be visible to all AWS endpoints", response.ProviderID, machine.Name)
 
 	operation := func() error {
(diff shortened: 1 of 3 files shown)


pkg/aws/core.go
```diff
@@ -223,18 +223,19 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
 
 	operation := func() error {
 		_, err := confirmInstanceByID(svc, *runResult.Instances[0].InstanceId)
 		return err
 	}
 
 	if err := retryWithExponentialBackOff(operation, maxElapsedTimeInBackoff); err != nil {
-		klog.V(3).Infof("Timed out waiting for VM %q to be visible to all AWS endpoints. Multiple VM backing machine obj %q might spawn, they will be orphan collected", response.ProviderID, machine.Name)
-		return nil, fmt.Errorf("creation of VM : %q Failed, timed out waiting for eventual consistency", response.ProviderID)
+		return nil, status.Error(codes.Internal, fmt.Sprintf("creation of VM : %q Failed, timed out waiting for eventual consistency. Multiple VMs backing machine obj might spawn, they will be orphan collected", response.ProviderID))
```
elankath, line 234: We should ideally wrap the source `err` here. Can omit the `:` in message. (For error messages, the convention is that `:` should separate error messages at diff nested levels)

himanshu-kun: yeah right. Actually there was no intent of nesting here . I have updated the error message , pls check.

elankath (approved):
thanks

PR #133, opened 2023-10-16
Moved Declaration of dynamic content from Makefile to .env file for AWS

piyuagr:
This PR updates Makefile content to import dynamic content from the .env file. 

Fixes -> https://github.com/gardener/machine-controller-manager/issues/846

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,18 +18,8 @@ IMAGE_REPOSITORY    := eu.gcr.io/gardener-project/gardener/machine-controller-ma
 IMAGE_TAG           := $(shell cat VERSION)
 PROVIDER_NAME       := AWS
 PROJECT_NAME        := gardener
-CONTROL_NAMESPACE	:=
-CONTROL_KUBECONFIG  := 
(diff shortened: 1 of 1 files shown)


himanshu-kun (approved):
/lgtm

Makefile
```diff
@@ -38,6 +28,7 @@ MACHINE_CONTROLLER_MANAGER_DEPLOYMENT_NAME := machine-controller-manager
 #########################################
 
 include hack/tools.mk
+include .env
```
himanshu-kun, line 31:
```suggestion
-include .env
```
this will ignore the .env file if it doesn't exist

himanshu-kun (changes requested):
Also I see logs like this printed, I remember this was added recently, kindly find where they are coming from and remove these logs
```
"TOOLS_BIN_DIR from tools.mk", <path-to-bin-dir>
"TOOLS_DIR from tools.mk", <path-to-tools-dir>
"PATH from tools.mk", <PATH>
```

PR #161, opened 2024-04-12
Add support for setting CpuOptions on AWS instances

AndreasBurger:
Adds support for setting CPU Options (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-specify-cpu-options.html) on instances.

**Release note**:

```feature user
It is now possible to specify CPU options for AWS instances.
```

### DIFF
--- a/pkg/aws/apis/validation/validation.go
+++ b/pkg/aws/apis/validation/validation.go
@@ -49,6 +49,7 @@ func ValidateAWSProviderSpec(spec *awsapi.AWSProviderSpec, secret *corev1.Secret
 	allErrs = append(allErrs, ValidateSecret(secret, field.NewPath("secretRef"))...)
 	allErrs = append(allErrs, validateSpecTags(spec.Tags, fldPath.Child("tags"))...)
 	allErrs = append(allErrs, validateInstanceMetadata(spec.InstanceMetadataOptions, fldPath.Child("instanceMetadata"))...)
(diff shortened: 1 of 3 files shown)


AndreasBurger:
Tests failing with `"no configuration has been provided, try setting KUBERNETES_MASTER environment variable"` but I don't think that's due to the changes proposed in this PR.

pkg/aws/apis/validation/validation.go
```diff
@@ -204,6 +205,27 @@ func validateInstanceMetadata(metadata *awsapi.InstanceMetadataOptions, fldPath
...
+		return allErrs
+	}
+
+	if cpuOptions.CoreCount == nil {
+		allErrs = append(allErrs, field.Required(fldPath.Child("coreCount"), "CoreCount is required"))
+	}
+
+	if cpuOptions.ThreadsPerCore == nil {
+		allErrs = append(allErrs, field.Required(fldPath.Child("threadsPerCore"), "ThreadsPerCore is required"))
+	}
```
kon-angelo, line 220: Just for my own understanding, is it working like that for the API: like you can only specify both or none ?

AndreasBurger: While I haven't actually tried this myself, the documentation states `The CPU options for the instance. Both the core count and threads per core must be specified in the request.`. Defaulting or some smart lookups do not work here as the "defaults" are different for each machine type, so we just require both.

kon-angelo (approved):
/lgtm

PR #173, opened 2024-09-05
Use `ProviderID` as a fallback for fetching the VM, `InitializeMachine` returns `Uninitialized` error code if VM is not found.

thiyyakat:

This PR introduces a fallback for fetching the VM instances associated with a machine. 

It also changes `InitializeMachine` to return error code `Uninitialized`, instead of `NotFound` when instances cannot be found for the machine.

Fixes part of https://github.com/gardener/machine-controller-manager/issues/933

The changes were manually tested by doing the following:

1. Returning an error from `getInstancesByTagsOrInstanceID()`, such that `instanceID` is fetched by the fallback `getInstanceByID()`. Found that Machine was created successfully and went into Running state.
2. returning `codes.NotFound` from `getMatchingInstancesForMachine()`. On triggering machine creation, initialization was tried in a loop after `shortRetry`.

**Release note**:

```other operator
Use `ProviderID` as a fallback for fetching the VM.
```

```other operator
`InitializeMachine` returns `Uninitialized` error code if VM is not found.
```

### DIFF
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -259,18 +259,21 @@ func (d *Driver) InitializeMachine(_ context.Context, request *driver.Initialize
 	if err != nil {
 		return nil, err
 	}
-	instances, err := d.getInstancesFromMachineName(request.Machine.Name, providerSpec, request.Secret)
+	instances, err := d.getMatchingInstancesForMachine(request.Machine, providerSpec, request.Secret)
 	if err != nil {
+		if isNotFoundError(err) {
+			//TODO check if level is correct, and if log is appropriate
+			klog.V(2).Infof("Could not retrieve instances for machine %s from provider", request.Machine.Name)
+			return nil, status.Error(codes.Uninitialized, err.Error())
+		}
 		return nil, err
 	}
 	targetInstance := instances[0]
 	providerID := encodeInstanceID(providerSpec.Region, *targetInstance.InstanceId)
-
 	svc, err := d.createSVC(request.Secret, providerSpec.Region)
 	if err != nil {
 		return nil, status.Error(codes.Uninitialized, err.Error())
 	}
-
 	// if SrcAnDstCheckEnabled is false then disable the SrcAndDestCheck on running NAT instance
 	if providerSpec.SrcAndDstChecksEnabled != nil && !*providerSpec.SrcAndDstChecksEnabled && *targetInstance.SourceDestCheck {
 		klog.V(3).Infof("Disabling SourceDestCheck on VM %q associated with machine %s", providerID, request.Machine.Name)
@@ -279,7 +282,6 @@ func (d *Driver) InitializeMachine(_ context.Context, request *driver.Initialize
 			return nil, status.Error(codes.Uninitialized, err.Error())
 		}
 	}
-
 	for i, netIf := range providerSpec.NetworkInterfaces {
 		for _, instanceNetIf := range targetInstance.NetworkInterfaces {
 			if netIf.Ipv6PrefixCount != nil && *instanceNetIf.Attachment.DeviceIndex == int64(i) {
@@ -368,7 +370,7 @@ func (d *Driver) DeleteMachine(_ context.Context, req *driver.DeleteMachineReque
 
 	} else {
 		// ProviderID doesn't exist, hence check for any existing machine and then delete if exists
-		instances, err = d.getInstancesFromMachineName(req.Machine.Name, providerSpec, req.Secret)
+		instances, err = d.getMatchingInstancesForMachine(req.Machine, providerSpec, req.Secret)
 		if err != nil {
 			errorStatus, ok := status.FromError(err)
 			if ok && errorStatus.Code() == codes.NotFound {
@@ -415,7 +417,7 @@ func (d *Driver) GetMachineStatus(_ context.Context, req *driver.GetMachineStatu
 		return nil, err
 	}
 
-	instances, err := d.getInstancesFromMachineName(req.Machine.Name, providerSpec, secret)
+	instances, err := d.getMatchingInstancesForMachine(req.Machine, providerSpec, secret)
 
 	if err != nil {
 		return nil, err
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -262,8 +262,7 @@ func (d *Driver) InitializeMachine(_ context.Context, request *driver.Initialize
 	instances, err := d.getMatchingInstancesForMachine(request.Machine, providerSpec, request.Secret)
 	if err != nil {
 		if isNotFoundError(err) {
-			//TODO check if level is correct, and if log is appropriate
-			klog.V(2).Infof("Could not retrieve instances for machine %s from provider", request.Machine.Name)
+			klog.Errorf("Could not get matching instance for uninitialized machine %q from provider: %s", request.Machine.Name, err)
 			return nil, status.Error(codes.Uninitialized, err.Error())
 		}
 		return nil, err
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -74,7 +74,7 @@ func (d *Driver) CreateMachine(_ context.Context, req *driver.CreateMachineReque
 
 	// Check if the MachineClass is for the supported cloud provider
 	if req.MachineClass.Provider != ProviderAWS {
-		err = fmt.Errorf("Requested for Provider '%s', we only support '%s'", req.MachineClass.Provider, ProviderAWS)
+		err = fmt.Errorf("requested for Provider '%s', we only support '%s'", req.MachineClass.Provider, ProviderAWS)
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
@@ -259,7 +259,11 @@ func (d *Driver) InitializeMachine(_ context.Context, request *driver.Initialize
 	if err != nil {
 		return nil, err
 	}
-	instances, err := d.getMatchingInstancesForMachine(request.Machine, providerSpec, request.Secret)
+	svc, err := d.createSVC(request.Secret, providerSpec.Region)
+	if err != nil {
+		return nil, status.Error(codes.Uninitialized, err.Error())
+	}
+	instances, err := d.getMatchingInstancesForMachine(request.Machine, svc, providerSpec.Tags)
 	if err != nil {
 		if isNotFoundError(err) {
 			klog.Errorf("Could not get matching instance for uninitialized machine %q from provider: %s", request.Machine.Name, err)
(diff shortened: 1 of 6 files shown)


gardener-robot-ci-2:
Thank you @thiyyakat for your contribution. Before I can start building your PR, a member of the organization must set the required label(s) {'reviewed/ok-to-test'}. Once started, you can check the build status in the PR checks section below.

pkg/aws/core.go
```diff
@@ -259,18 +259,20 @@ func (d *Driver) InitializeMachine(_ context.Context, request *driver.Initialize
 	if err != nil {
 		return nil, err
 	}
-	instances, err := d.getInstancesFromMachineName(request.Machine.Name, providerSpec, request.Secret)
+	instances, err := d.getMatchingInstancesForMachine(request.Machine, providerSpec, request.Secret)
```
unmarshall, line 262: can we just pass `providerSpec.Tags` to `getMachineInstancesByTagsAndStatus` function? Inside the function you can extract cluster name and node role.

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+		}
+	}
+	instances, err = getMachineInstancesByTagsAndStatus(svc, machine.Name, clusterName, nodeRole)
+	if err != nil {
+		return nil, err
 	}
 	if len(instances) == 0 {
-		errMessage := "AWS plugin is returning no VM instances backing this machine object"
-		return nil, status.Error(codes.NotFound, errMessage)
+		//if getMachineInstancesByTagsAndStatus returns an error, try fetching matching instances using ProviderID
```
unmarshall, line 151: This comment is misleading. If there was an error returned by `getMachineInstancesByTagsAndStatus` then you will not even come here but return much earlier. So change this comment.

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+			clusterName = key
+		} else if strings.Contains(key, "kubernetes.io/role/") {
+			nodeRole = key
+		}
+	}
+	instances, err = getMachineInstancesByTagsAndStatus(svc, machine.Name, clusterName, nodeRole)
+	if err != nil {
+		return nil, err
 	}
 	if len(instances) == 0 {
```
unmarshall, line 150: can you add a log stating that no VM was found using tags/status, now looking for VM using the providerID? This will help in diagnosing cases that we ran into where no VM was found when looking for a VM using tags + status.

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+		return instances, nil
 	}
+}
 
-	for _, reservation := range runResult.Reservations {
-		instances = append(instances, reservation.Instances...)
+// getMatchingInstancesForMachine extracts AWS Instance object for a given machine
+func (d *Driver) getMatchingInstancesForMachine(machine *v1alpha1.Machine, providerSpec *api.AWSProviderSpec, secret *corev1.Secret) ([]*ec2.Instance, error) {
+	var (
+		clusterName string
```
unmarshall, line 131: considering that you will move the extraction of clusterName and nodeRole elsewhere there is no reason to define these variables and you can just use short-assign.

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+			instances = append(instances, reservation.Instances...)
+		}
+		return instances, nil
 	}
+}
 
-	for _, reservation := range runResult.Reservations {
-		instances = append(instances, reservation.Instances...)
+// getMatchingInstancesForMachine extracts AWS Instance object for a given machine
+func (d *Driver) getMatchingInstancesForMachine(machine *v1alpha1.Machine, providerSpec *api.AWSProviderSpec, secret *corev1.Secret) ([]*ec2.Instance, error) {
```
unmarshall, line 129: to prevent shadowing you can just define names for the return types so the function definition will become:
```go
func (d *Driver) getMatchingInstancesForMachine(machine *v1alpha1.Machine, providerSpec *api.AWSProviderSpec, secret *corev1.Secret) (instances []*ec2.Instance, err error) {
}
```

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+		}
+		runResult, err := getInstanceByID(svc, instanceID)
+		if err != nil {
+			return nil, err
+		}
+		for _, reservation := range runResult.Reservations {
+			instances = append(instances, reservation.Instances...)
+		}
+		if len(instances) == 0 {
+			errMessage := "AWS plugin is returning no VM instances backing this machine object"
```
unmarshall, line 167: what is `AWS plugin` mean here?

pkg/aws/core_util.go
```diff
@@ -131,34 +111,87 @@ func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *a
...
+	}
+	instances, err = getMachineInstancesByTagsAndStatus(svc, machine.Name, clusterName, nodeRole)
+	if err != nil {
+		return nil, err
 	}
 	if len(instances) == 0 {
-		errMessage := "AWS plugin is returning no VM instances backing this machine object"
-		return nil, status.Error(codes.NotFound, errMessage)
+		//if getMachineInstancesByTagsAndStatus returns an error, try fetching matching instances using ProviderID
+		if machine.Spec.ProviderID == "" {
```
unmarshall, line 152: maybe you can also move the logic of extraction of instanceID from providerID into `getInstanceByID`

pkg/aws/core.go
```diff
@@ -368,7 +369,7 @@ func (d *Driver) DeleteMachine(_ context.Context, req *driver.DeleteMachineReque
 
 	} else {
 		// ProviderID doesn't exist, hence check for any existing machine and then delete if exists
-		instances, err = d.getInstancesFromMachineName(req.Machine.Name, providerSpec, req.Secret)
+		instances, err = d.getMatchingInstancesForMachine(req.Machine, providerSpec, req.Secret)
```
unmarshall, line 372: In the `If` condition you have already checked that the providerID is empty. Therefore just calling `getMachineInstancesByTagsAndStatus` should be sufficient, right? This also prevents recreating the service which you have already created once.

pkg/aws/core.go
```diff
@@ -415,7 +416,7 @@ func (d *Driver) GetMachineStatus(_ context.Context, req *driver.GetMachineStatu
 		return nil, err
 	}
 
-	instances, err := d.getInstancesFromMachineName(req.Machine.Name, providerSpec, secret)
+	instances, err := d.getMatchingInstancesForMachine(req.Machine, providerSpec, secret)
```
unmarshall, line 419: In the code below:
```go
	if err != nil {
		return nil, err
	} else if len(instances) > 1 {
		instanceIDs := []string{}
		for _, instance := range instances {
			instanceIDs = append(instanceIDs, *instance.InstanceId)
		}

		errMessage := fmt.Sprintf("AWS plugin is returning multiple VM instances backing this machine object. IDs for all backing VMs - %v ", instanceIDs)
		return nil, status.Error(codes.OutOfRange, errMessage)
	}
```
Can you change the following:
```go
 instanceIDs := []string{}
```
change the above to:
```go
  instanceIDs := make([]string, 0, len(instances))
```

pkg/aws/util.go
```diff
@@ -45,6 +48,15 @@ func (d *Driver) createSVC(secret *corev1.Secret, region string) (ec2iface.EC2AP
 	return svc, nil
 }
 
+// Function returns true only if error code equals codes.NotFound
+func isNotFoundError(err error) bool {
```
unmarshall, line 52: can you check places where we extract the error code and compare it against `codes.NotFound` and in those places use this function instead?
One place on quick search is `core_util.go/terminateInstance`

unmarshall (changes requested):

pkg/aws/core_util.go
thiyyakat: AWS plugin here means the EC2 service that we create for describing the instances.

pkg/aws/core_util.go
thiyyakat: `getInstanceByID` is used in `confirmInstanceByID` where we directly pass the InstanceID.  `confirmInstanceByID` is used in `CreateMachine` where we generate the providerID using the instanceID, so it does not make sense to pass the providerID just to extract the instanceID back from it.

pkg/aws/util.go
thiyyakat: We cannot change `terminateInstance` because here we extract the MCM error code from the provider error and then compare it with `codes.NotFound`. We have used this function in one more place, where we found the usage was possible.

pkg/aws/core_util.go
unmarshall: Ok this is a bit unclear. Can we just say `No VM instances returned by AWS EC2 service for the machine object`

pkg/aws/core_util.go
unmarshall: Ok

pkg/aws/core.go
unmarshall: There was actually no reason to pass the entire providerSpec.  Instead only pass the `service`. At present in `InitializeMachine` you are constructing it twice which is not really required.

pkg/aws/core_util.go
```diff
@@ -79,27 +79,18 @@ func disableSrcAndDestCheck(svc ec2iface.EC2API, instanceID *string) error {
 	return nil
 }
 
-// getInstancesFromMachineName extracts AWS Instance object from given machine name
-func (d *Driver) getInstancesFromMachineName(machineName string, providerSpec *api.AWSProviderSpec, secret *corev1.Secret) ([]*ec2.Instance, error) {
+func getMachineInstancesByTagsAndStatus(svc ec2iface.EC2API, machineName string, providerSpecTags map[string]string) ([]*ec2.Instance, error) {
```
unmarshall, line 82: I know that this is an existing piece of code but nevertheless i have a question.
If the tags for cluster and node role are missing then what will describeInstance return?
Will will the impact?

thiyyakat: If the tags for cluster or nodeRole are missing, then `getMachineInstancesByTagsAndStatus` does not return an error, but returns a nil slice of instances. So we will try to fetch the instance by its ID.

thiyyakat: We set the cluster and node role to empty strings. This results in an empty slice, and the fallback is tried.

unmarshall: ok then we are good. Thanks for testing this.

unmarshall (approved):
/lgtm

PR #174, opened 2024-09-10
Set PrimaryIpv6 flag for instances with IPv6 addresses.

axel7born:
This is needed to add an instance with an IPv6 address as target to a loadbalancer.

Fixes #

**Release note**:

```other operator
Set PrimaryIpv6 flag for instances with IPv6 addresses.
```


rishabh-11 (approved):
/lgtm

PR #184, opened 2025-01-23
Update `pipeline_integration_tests` to run tests on a control cluster

aaronfern:
This PR does the following
1. Updated the `pipeline_integration_tests` script to now use a control plane as a control cluster to deploy all objects needed for the tests
2. Update the `MCM` version to `v0.56.0`

Fixes partially gardener/machine-controller-manager#960

**Release note**:

```other operator
vendored MCM version has been upgraded to v0.56.0
```
```other operator
pipeline_integration_test now uses the control plane of a cluster to deploy objects part of its tests
```


.ci/pipeline_integration_test
```diff
@@ -21,66 +22,19 @@ KUBECONFIG_EXPIRY=10800
 declare CONTROL_KUBECONFIG=${SOURCE_PATH}/dev/control_kubeconfig.yaml
 declare TARGET_KUBECONFIG=${SOURCE_PATH}/dev/target_kubeconfig.yaml
 declare TEST_CREDS_SECRET=test-mc-secret
+declare IS_CONTROL_CLUSTER_SEED=true
+declare PROJECT=mcm-ci
+declare CONTROL_CLUSTER_NAMESPACE=shoot--"${PROJECT}"--"${TEST_SHOOT}"
+declate GNA_SECRET_NAME
```
takoverflow, line 28: Typo: should be `declare`

.ci/pipeline_integration_test
```diff
@@ -105,69 +56,24 @@ function fetch_oot_cluster_kubeconfig() {
...
-kind: Secret
-metadata:
-  name: $TEST_CREDS_SECRET
-  namespace: default
-type: Opaque
-EOF
-)
-    echo "$secret_yaml" | kubectl --kubeconfig=$CONTROL_KUBECONFIG apply -f -
+function fetch_gna_secret_name(){
+    MCC_NAME=$(kubectl --kubeconfig=dev/control_kubeconfig.yaml get mcc -n $CONTROL_CLUSTER_NAMESPACE -o custom-columns=Name:.metadata.name --no-headers | grep shoot--mcm-ci)
```
elankath, line 69: we don't need multiple kubectl commands here. We can download the machine class yamls into `dev/v1machineclass_converted.yaml` as a single kubectl command and then  get the get the first element of the array and get the `MCC_NAME` and  `GNA_SECRET_NAME`

elankath (approved):
OK for now. but we need to refix the MCM to be able to work in shoot cluster. unhappy that we are forced to do this.

.ci/pipeline_integration_test
aaronfern: Updated to use only one kubectl command

PR #190, opened 2025-03-28
Add API metrics for instance methods

### LINKED ISSUE #156: Record Metrics for AWS API calls
In version [v0.50.0](https://github.com/gardener/machine-controller-manager/releases/tag/v0.50.0) of MCM new metrics were added to record cloud provider api request durations and failures. It is for the provider's to implement the logic of updating these metrics at the appropriate points in the code base based on their driver activity with the provider.

As a part of this issue, the provider needs to record `requests_failed_total`, `requests_total` and `api_request_duration_seconds` metrics. Refer `mcm-provider-az` repo to see how they are implemented for Azure API calls.

**Why is this needed**:
Corresponding MCM issue :- https://github.com/gardener/machine-controller-manager/issues/872

takoverflow:

Fixes #156

**Release note**:

```other operator

```

### DIFF
--- a/pkg/instrument/instrument.go
+++ b/pkg/instrument/instrument.go
@@ -6,14 +6,46 @@ package instrument
 
 import (
 	"errors"
+	"strconv"
+	"time"
+
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/machinecodes/codes"
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/machinecodes/status"
 	"github.com/gardener/machine-controller-manager/pkg/util/provider/metrics"
-	"strconv"
-	"time"
 )
 
 const prometheusProviderLabelValue = "aws"
+
+// RecordAwsAPIMetric records a prometheus metric for AWS API calls.
+// * If there is an error then it will increment the APIFailedRequestCount counter vec metric.
+// * If the AWS API call is successful then it will record 2 metrics:
+//   - It will increment APIRequestCount counter vec metric.
+//   - It will compute the time taken for API call completion and record it.
+//
+// NOTE: If this function is called via `defer` then please keep in mind that parameters passed to defer are evaluated at the time of definition.
+// So if you have an error that is computed later in the function then ensure that you use named return parameters.
+func RecordAwsAPIMetric(err error, awsServiceName string, invocationTime time.Time) {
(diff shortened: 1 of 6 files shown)


takoverflow:
Ran `prometheus` and `mcm` locally to verify the capturing of the API metrics
image

pkg/instrument/instrument.go
```diff
@@ -42,6 +74,15 @@ func RecordDriverAPIMetric(err error, operation string, invocationTime time.Time
 	).Observe(elapsed.Seconds())
 }
 
+// AwsAPIMetricRecorderFn returns a function that can be used to record a prometheus metric for AWS API calls.
+// NOTE: a pointer to an error (which itself is a fat interface pointer) is necessary to enable the callers of this function to enclose this call into a `defer` statement.
+func AwsAPIMetricRecorderFn(awsServiceName string, err *error) func() {
+	invocationTime := time.Now()
```
aaronfern, line 80: I think I see a problem here. `InvocationTime` here is only evaluated when the caller api func returns, which would result in incorrect `elapsed` time being recorded as part of the metric.
Ideally `invocationTime` should be passed as a parameter to this function. In that case, the value is evaluated at the start of the api function which results in accurate `elapsed` time

takoverflow: As discussed offline, the `defer RecorderFn(...)()` call defers the execution of the function returned by the `RecorderFn` but not the contents of the `RecorderFn` itself hence the `invocationTime` would be evaluated at the start of the `API` function whose duration is being measured.

<details><summary> Example</summary>

```go
func Record(inTime time.Time) {
    time.Sleep(5 * time.Second)
    elapsed := time.Since(inTime)
    fmt.Printf("Elapsed: %v\n", elapsed)
}

func APIRecorderFn() func() {
    inTime := time.Now()
    return func() {
        Record(inTime)
    }
}

func measureDelay() {
    defer APIRecorderFn()()
    time.Sleep(3 * time.Second)
}

func main() {
    measureDelay()
}
```

Returns `Elapsed: 8s` thus confirming the same
</details>

aaronfern: Thanks for the explanation!

aaronfern (approved):

PR #197, opened 2025-07-02
Set region when creating session

dimityrmirchev:
If the region is not set the retrieval of credentials might fail with the following error:

```
WebIdentityErr: failed to retrieve credentials
caused by: InvalidIdentityToken: No OpenIDConnect provider found in your account for https://issuer.example
        status code: 400, request id: <id-of-request>
```

Fixes #

**Release note**:

```bugfix user
A bug causing failure in retrieval of credentials when using workload identity due to missing region parameter was fixed.
```

### DIFF
--- a/pkg/spi/aws.go
+++ b/pkg/spi/aws.go
@@ -29,7 +29,7 @@ func (ms *PluginSPIImpl) NewSession(secret *corev1.Secret, region string) (*sess
 	}
 
 	if workloadIdentityTokenFile, ok := secret.Data["workloadIdentityTokenFile"]; ok {
-		sess, err := session.NewSession()
+		sess, err := session.NewSession(aws.NewConfig().WithRegion(region))
 		if err != nil {
 			return nil, err
 		}


pkg/spi/aws.go
```diff
@@ -29,7 +29,7 @@ func (ms *PluginSPIImpl) NewSession(secret *corev1.Secret, region string) (*sess
 	}
 
 	if workloadIdentityTokenFile, ok := secret.Data["workloadIdentityTokenFile"]; ok {
-		sess, err := session.NewSession()
+		sess, err := session.NewSession(aws.NewConfig().WithRegion(region))
```
aaronfern, line 32: Is there an advantage in creating a new aws.Config{} object over using the variable `config` defined at the start of this function?

aaronfern (commented):
Thanks for the PR @dimityrmirchev 
Just one question, thanks!

pkg/spi/aws.go
dimityrmirchev: I also thought of that but decided to use a separate config since we return a new session at the end anyways https://github.com/gardener/machine-controller-manager-provider-aws/blob/9f7238e49d935544416d2e9a542aca9014751551/pkg/spi/aws.go#L48. I was thinking that I cannot be sure if the `session.NewSession` function would cause any side effect to the passed config so I decided to pass a new one to it.

aaronfern: Okay, fair enough
In that case, this is fine

PR #244, opened 2026-04-13
Upgrade golang to 1.26.2

takoverflow:
Bumps go version to `1.26.2`
- Migrate to golangci-lint v2
- Bump k8s api from `v0.31.0` to `v0.34.0`

Part of gardener/machine-controller-manager#1081

**Release note**:

```other operator
Golang version used is now upgraded to `1.26.2`
```


.golangci.yaml
```diff
@@ -1,49 +1,60 @@
+version: "2"
 run:
   concurrency: 4
-  timeout: 10m
```
thiyyakat, line 3: Is it fine that this timeout was removed?

thiyyakat (commented):
Thanks for the PR. I just had one question.

.golangci.yaml
takoverflow: As per https://golangci-lint.run/docs/product/migration-guide/#command-migrate

> The migrate command enforces the following default values:
run.timeout: the existing value is ignored because, in v2, there is no timeout by default.

thiyyakat (approved):

PR #247, opened 2026-05-05
Add EFA network interface, placement, and ML capacity block support

### LINKED ISSUE #246: Add EFA network interface, placement group, and ML capacity block support

Support for EFA network interfaces, placement groups, and ML capacity blocks in the MCM provider:
- `AWSNetworkInterfaceSpec`: Add InterfaceType, NetworkCardIndex, DeviceIndex, PrimaryIpv6 fields
- `AWSProviderSpec`: Add Placement field (groupId, tenancy, hostId, partitionNumber, affinity)
- `AWSCapacityReservationTargetSpec`: Add IsMLCapacityBlock field
- Fix `disableSrcAndDestCheck` to skip EFA/efa-only interfaces
- Fix `GetMachineStatus` to check SourceDestCheck per-interface (not instance-level)

Related: gardener/gardener-extension-provider-aws#1790

**Why is this needed**:

GPU instances with EFA require multiple network interfaces with specific types (efa, efa-only) and network card indices. The current MCM provider only supports basic ENI configuration. Additionally, EFA interfaces don't support SourceDestCheck modification, causing machines to get stuck in Uninitialized state.

shreyas-s-rao:
/area control-plane
/kind api-change
/platform aws

Adds support for EFA network interfaces, placement groups, and ML capacity blocks:
- Add InterfaceType, NetworkCardIndex, DeviceIndex, PrimaryIpv6 to AWSNetworkInterfaceSpec
- Add AWSPlacementSpec (groupId, tenancy, hostId, partitionNumber, affinity) to AWSProviderSpec
- Add InstanceMarketOptions.MarketType with currently valid values `spot`, `capacity-block`, `interruptible-capacity-reservation`
- Deprecate the overloaded SpotPrice in favour of using `InstanceMarketOptions.MarketType: spot` to explicitly specify using spot instances. Additionally, actual spot price feature is not used by extension-provider-aws as of today.
- Fix disableSrcAndDestCheck to skip EFA/efa-only interfaces (per-interface instead of per-instance)
- Fix GetMachineStatus to check SourceDestCheck per-interface (prevents Uninitialized loop on EFA instances)
- Spec-based placement with annotation fallback for backward compatibility
- Validation for new fields

Fixes #246

- GetMachineStatus SourceDestCheck fix is critical: without it, EFA machines get stuck in Uninitialized state

**Locally tested cases**:

- [x] g4dn.12xlarge machine Running with EFA
- [x] p4d.24xlarge machines Running with 4 EFA NICs
- [x] Placement group applied via providerSpec
- [x] SourceDestCheck per-interface (no Uninitialized loop)
- [x] Validation: invalid interfaceType rejected
- [x] Validation: negative deviceIndex rejected
- [x] Validation: negative networkCardIndex rejected
- [x] Validation: isMLCapacityBlock without ID rejected
- [x] Validation: invalid placement tenancy rejected
- [x] Validation: hostId without tenancy host rejected

**Release note**:

```feature user
Add support for EFA network interfaces (InterfaceType, NetworkCardIndex, DeviceIndex) in EC2 instance creation
```

```feature user
Add placement group, tenancy, and dedicated host configuration support via providerSpec
```

```feature user
Add `InstanceMarketOptions.MarketType` to specify market-type options such as `spot`, `capacity-block` (ML capacity block reservations), etc.
```

```noteworthy user
`SpotPrice` is now deprecated, and will be removed in a future release. Please use `InstanceMarketOptions.MarketType: spot` instead.
```

```bugfix user
Fix SourceDestCheck validation for EFA instances causing machines to get stuck in Uninitialized state
```

### DIFF
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -100,6 +100,9 @@ type AWSProviderSpec struct {
 
 	// CPUOptions contains detailed configuration for the number of cores and threads for the instance.
 	CPUOptions *CPUOptions `json:"cpuOptions,omitempty"`
+
+	// Placement contains placement configuration for the instance (placement groups, tenancy, dedicated hosts).
+	Placement *AWSPlacementSpec `json:"placement,omitempty"`
 }
 
 // AWSBlockDeviceMappingSpec stores info about AWS block device mappings
@@ -140,6 +143,10 @@ type AWSCapacityReservationTargetSpec struct {
 
 	// CapacityReservationResourceGroupArn The ARN of the Capacity Reservation in which to run the instance.
 	CapacityReservationResourceGroupArn *string `json:"capacityReservationResourceGroupArn,omitempty"`
+
+	// IsMLCapacityBlock indicates this reservation is an ML capacity block.
+	// Sets InstanceMarketOptions.MarketType to "capacity-block".
+	IsMLCapacityBlock *bool `json:"isMLCapacityBlock,omitempty"`
 }
 
 // AWSEbsBlockDeviceSpec describes a block device for an EBS volume.
@@ -247,6 +254,33 @@ type AWSNetworkInterfaceSpec struct {
 	// The ID of the subnet associated with the network string. Applies only if
 	// creating a network interface when launching an machine.
 	SubnetID string `json:"subnetID,omitempty"`
+
+	// InterfaceType is the type of network interface.
+	// Valid values: "interface" (default), "efa", "efa-only".
+	InterfaceType string `json:"interfaceType,omitempty"`
+
+	// NetworkCardIndex is the index of the network card for this interface.
+	NetworkCardIndex *int32 `json:"networkCardIndex,omitempty"`
+
+	// DeviceIndex is the device index for this network interface.
+	DeviceIndex *int32 `json:"deviceIndex,omitempty"`
+
+	// PrimaryIpv6 indicates whether the first IPv6 address is the primary IPv6 address.
+	PrimaryIpv6 *bool `json:"primaryIpv6,omitempty"`
+}
+
+// AWSPlacementSpec contains placement configuration for an EC2 instance.
+type AWSPlacementSpec struct {
+	// GroupID is the ID of the placement group.
+	GroupID *string `json:"groupId,omitempty"`
+	// Tenancy is the tenancy of the instance. Valid values: "default", "dedicated", "host".
+	Tenancy *string `json:"tenancy,omitempty"`
+	// HostID is the ID of the Dedicated Host.
+	HostID *string `json:"hostId,omitempty"`
+	// PartitionNumber is the partition number for the instance.
+	PartitionNumber *int64 `json:"partitionNumber,omitempty"`
+	// Affinity is the affinity setting. Valid values: "default", "host".
+	Affinity *string `json:"affinity,omitempty"`
 }
 
 const (
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -85,8 +85,9 @@ type AWSProviderSpec struct {
 	// Region contains the AWS region for the machine
 	Region string `json:"region,omitempty"`
 
-	// SpotPrice is an optional field that if set specifies to use spot instances
-	// When set to "" there is no maxPrice else, specifies the maxPrice
+	// SpotPrice is an optional field that if set specifies to use spot instances.
+	// When set to "" there is no maxPrice, else specifies the maxPrice.
+	// Deprecated: Use InstanceMarketOptions with MarketType "spot" instead.
 	SpotPrice *string `json:"spotPrice,omitempty"`
 
 	// If set to false, source and destination checks are disabled, default is true
@@ -103,6 +104,10 @@ type AWSProviderSpec struct {
 
 	// Placement contains placement configuration for the instance (placement groups, tenancy, dedicated hosts).
 	Placement *AWSPlacementSpec `json:"placement,omitempty"`
+
+	// InstanceMarketOptions configures the instance market type.
+	// If not specified, on-demand instances are launched.
+	InstanceMarketOptions *AWSInstanceMarketOptions `json:"instanceMarketOptions,omitempty"`
 }
 
 // AWSBlockDeviceMappingSpec stores info about AWS block device mappings
@@ -143,10 +148,13 @@ type AWSCapacityReservationTargetSpec struct {
 
 	// CapacityReservationResourceGroupArn The ARN of the Capacity Reservation in which to run the instance.
 	CapacityReservationResourceGroupArn *string `json:"capacityReservationResourceGroupArn,omitempty"`
+}
 
-	// IsMLCapacityBlock indicates this reservation is an ML capacity block.
-	// Sets InstanceMarketOptions.MarketType to "capacity-block".
-	IsMLCapacityBlock *bool `json:"isMLCapacityBlock,omitempty"`
+// AWSInstanceMarketOptions configures the instance market type.
+type AWSInstanceMarketOptions struct {
+	// MarketType is the market type for the instance.
+	// Supported values: "spot", "capacity-block", "interruptible-capacity-reservation".
+	MarketType string `json:"marketType"`
 }
 
 // AWSEbsBlockDeviceSpec describes a block device for an EBS volume.
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -280,7 +280,7 @@ type AWSPlacementSpec struct {
 	// HostID is the ID of the Dedicated Host.
 	HostID *string `json:"hostId,omitempty"`
 	// PartitionNumber is the partition number for the instance.
-	PartitionNumber *int64 `json:"partitionNumber,omitempty"`
+	PartitionNumber *int32 `json:"partitionNumber,omitempty"`
 	// Affinity is the affinity setting. Valid values: "default", "host".
 	Affinity *string `json:"affinity,omitempty"`
 }
--- a/pkg/aws/apis/aws_provider_spec.go
+++ b/pkg/aws/apis/aws_provider_spec.go
@@ -256,8 +256,10 @@ type AWSNetworkInterfaceSpec struct {
 	SubnetID string `json:"subnetID,omitempty"`
 
 	// InterfaceType is the type of network interface.
-	// Valid values: "interface" (default), "efa", "efa-only".
-	InterfaceType string `json:"interfaceType,omitempty"`
+	// Currently valid values for RunInstances: "interface", "efa", "efa-only".
+	// See https://github.com/aws/aws-sdk-go-v2/blob/service/ec2/v1.279.0/service/ec2/types/types.go#L9181
+	// If not specified, defaults to "interface".
+	InterfaceType *string `json:"interfaceType,omitempty"`
 
 	// NetworkCardIndex is the index of the network card for this interface.
 	NetworkCardIndex *int32 `json:"networkCardIndex,omitempty"`
--- a/pkg/aws/apis/validation/validation.go
+++ b/pkg/aws/apis/validation/validation.go
@@ -42,6 +42,7 @@ func ValidateAWSProviderSpec(spec *awsapi.AWSProviderSpec, secret *corev1.Secret
 	allErrs = append(allErrs, validateBlockDevices(spec.BlockDevices, fldPath.Child("blockDevices"))...)
 	allErrs = append(allErrs, validateCapacityReservations(spec.CapacityReservationTarget, fldPath.Child("capacityReservation"))...)
 	allErrs = append(allErrs, validateNetworkInterfaces(spec.NetworkInterfaces, fldPath.Child("networkInterfaces"))...)
+	allErrs = append(allErrs, validatePlacement(spec.Placement, fldPath.Child("placement"))...)
 	allErrs = append(allErrs, ValidateSecret(secret, field.NewPath("secretRef"))...)
 	allErrs = append(allErrs, validateSpecTags(spec.Tags, fldPath.Child("tags"))...)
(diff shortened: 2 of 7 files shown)


pkg/aws/apis/aws_provider_spec.go
```diff
@@ -140,6 +143,10 @@ type AWSCapacityReservationTargetSpec struct {
 
 	// CapacityReservationResourceGroupArn The ARN of the Capacity Reservation in which to run the instance.
 	CapacityReservationResourceGroupArn *string `json:"capacityReservationResourceGroupArn,omitempty"`
+
+	// IsMLCapacityBlock indicates this reservation is an ML capacity block.
+	// Sets InstanceMarketOptions.MarketType to "capacity-block".
+	IsMLCapacityBlock *bool `json:"isMLCapacityBlock,omitempty"`
```
elankath: From what I understand, the AWS Provider `RunInstancesInput.InstanceMarketOptions.MarketType` which can be a range of enum variants. So, in future if we introduce support for another enum variant type, then we have to introduce yet another bool field. Don't you think for future compatibility and growth, it is better to have `[]MarkedType` somewhere else. Please note that even in the current code, `MarketType` is not part of `ec2types.CapacityReservationSpecification` which is set under `RunInstancesInput.CapacityReservationSpecification` but separately under `RunInstancesInput.InstanceMarketOptions.MarketType`. Ie positioning of this field appears wrong here.

pkg/aws/apis/aws_provider_spec.go
```diff
@@ -247,6 +254,33 @@ type AWSNetworkInterfaceSpec struct {
 	// The ID of the subnet associated with the network string. Applies only if
 	// creating a network interface when launching an machine.
 	SubnetID string `json:"subnetID,omitempty"`
+
+	// InterfaceType is the type of network interface.
+	// Valid values: "interface" (default), "efa", "efa-only".
+	InterfaceType string `json:"interfaceType,omitempty"`
```
elankath: Why use string here and not `*string` ? since it would match `InstanceNetworkInterfaceSpecification.InterfaceType` which is `*string`.

pkg/aws/core.go
```diff
@@ -138,20 +138,37 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
 		}
 
+		if netIf.DeviceIndex != nil {
+			spec.DeviceIndex = netIf.DeviceIndex
+		} else {
+			spec.DeviceIndex = aws.Int32(int32(i)) // #nosec: G115 -- index will not exceed int32 limits
+		}
+
+		if netIf.NetworkCardIndex != nil {
+			spec.NetworkCardIndex = netIf.NetworkCardIndex
```
elankath: `nil` check not needed. Both are `*int32`. can directly assign.

pkg/aws/core.go
```diff
@@ -138,20 +138,37 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+		}
+
+		if netIf.InterfaceType != "" {
+			spec.InterfaceType = ptr.To(netIf.InterfaceType)
```
elankath: No need for `if` check if `spec.InterfaceType` was `*string`.

pkg/aws/apis/aws_provider_spec.go
shreyas-s-rao: Understood. But what should happen to `AWSProviderSpec.SpotPrice`? Specifying it to a non-nil value sets `RunInstancesInput.InstanceMarketOptions.MarketType` to `MarketTypeSpot`. So even if I add a new field like `AWSProviderSpec.MarketType` that can take any arbitrary value (a bit risky without validation IMO), it would still clash with options like `SpotPrice` right? Do you think there's a way to harmonize this?

pkg/aws/apis/aws_provider_spec.go
shreyas-s-rao: Because the API package does not import the AWS SDK at all, and I didn't want to add that heavy dependency to it, keeping in mind the possibility of creating a separate API go module like many other gardener repos are adopting. Also, the struct fields are JSON-deserialized from the MachineClass providerSpec, so it would be better if they are simple types.

That said, I've switched to `*string` for this field in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/commits/e4292e8089aa5054a298cc5aff73e25987e4840c

pkg/aws/core.go
shreyas-s-rao: Done in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/commits/e4292e8089aa5054a298cc5aff73e25987e4840c

pkg/aws/core.go
shreyas-s-rao: Good catch. I had used `string` type in extension-provider-aws `WorkerConfig` for `NetworkInterface.Type`, hence needed this additional validation here. But that field itself should have been a `*string`, since it's an optional field. I've made the change now in PRs of both repos. PTAL.
Fixed in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/commits/e4292e8089aa5054a298cc5aff73e25987e4840c

shreyas-s-rao:
@elankath thanks for your review. I've addressed your comments. PTAL, thanks.

pkg/aws/apis/aws_provider_spec.go
takoverflow: AFAIK currently spot instances support is not there.

I see that this particular field was introduced a long time back

https://github.com/gardener/machine-controller-manager/issues/27#issuecomment-673690704

And still there's no spot instances support in `gardener` afaik. So if required, the `SpotPrice` field can be changed/migrated without consequences. If it leads to a cleaner API, it would be better to consolidate information related to `MarketType`s.

(Please correct me if I'm wrong @elankath)

pkg/aws/apis/aws_provider_spec.go
takoverflow: It does though indirectly, because the `validation` of the provider spec inside `apis` does import AWS SDK.

https://github.com/gardener/machine-controller-manager-provider-aws/blob/914d01674c9ed79bb21b7964a0eb19f759ecc83c/pkg/aws/apis/validation/validation.go#L14

As for the question of having a separate API module for MCM, I would be interested in understanding the use-case of such a module, is it just for importing the required data for `gardener-extension-provider-aws`?

pkg/aws/apis/validation/validation.go
```diff
@@ -143,11 +144,22 @@ func validateCapacityReservations(capacityReservation *awsapi.AWSCapacityReserva
...
+
+		if capacityReservation.IsMLCapacityBlock != nil && *capacityReservation.IsMLCapacityBlock && capacityReservation.CapacityReservationID == nil {
+			allErrs = append(allErrs, field.Forbidden(fldPath.Child("isMLCapacityBlock"), "isMLCapacityBlock can only be set when capacityReservationId is specified"))
+		}
 	}
 
 	return allErrs
 }
 
+var validNetworkInterfaceTypes = []string{
```
takoverflow, line 153: All of this casting would be avoided if `AWSNetworkInterfaceSpec.InterfaceType` was defined to be `NetworkInterfaceType` instead?

Then this would just be a `[]NetworkInterfaceType`.

pkg/aws/apis/validation/validation.go
```diff
@@ -287,3 +313,34 @@ func validateStringValues(fld *field.Path, s string, accepted []string) field.Er
...
+	allErrs := field.ErrorList{}
+	if placement == nil {
+		return allErrs
+	}
+
+	if placement.Tenancy != nil {
+		validTenancies := []string{"default", "dedicated", "host"}
+		if !slices.Contains(validTenancies, *placement.Tenancy) {
+			allErrs = append(allErrs, field.NotSupported(fldPath.Child("tenancy"), *placement.Tenancy, validTenancies))
+		}
```
takoverflow, line 324: Again, if Tenancy was defined to be `*ec2Types.Tenancy`, this would've been simpler and wouldn't need updating if `Tenancy` types are expanded in the future

```suggestion
		validTenancies := *placement.Tenancy.Values()
		if !slices.Contains(validTenancies, *placement.Tenancy) {
			allErrs = append(allErrs, field.NotSupported(fldPath.Child("tenancy"), *placement.Tenancy, validTenancies))
		}
```

This could similarly be extended to other fields.

I don't see anyone using standalone `MCM-provider-AWS` api without depending on the upstream AWS-SDK anyways, WDYT @shreyas-s-rao

pkg/aws/core.go
```diff
@@ -214,14 +226,41 @@ func (d *Driver) CreateMachine(ctx context.Context, req *driver.CreateMachineReq
...
+			placement.GroupId = providerSpec.Placement.GroupID
+		}
+		if providerSpec.Placement.Tenancy != nil {
+			placement.Tenancy = ec2types.Tenancy(*providerSpec.Placement.Tenancy)
+		}
+		if providerSpec.Placement.HostID != nil {
+			placement.HostId = providerSpec.Placement.HostID
+		}
+		if providerSpec.Placement.PartitionNumber != nil {
+			placement.PartitionNumber = aws.Int32(int32(*providerSpec.Placement.PartitionNumber)) // #nosec: G115 -- partition number will not exceed int32 limits
```
takoverflow: Why not define `PartitionNumber` to be `*int32` instead? Avoiding all the casting and pointer construction.

pkg/aws/core_util.go
```diff
@@ -61,22 +61,46 @@ func decodeProviderSpecAndSecret(machineClass *v1alpha1.MachineClass, secret *co
 	return providerSpec, nil
 }
 
-// disableSrcAndDestCheck disbales the SrcAndDestCheck for NAT instances
```
takoverflow, line 64: Can the doc-string be updated to inform about the EFA interfaces skipping rather than removed?

pkg/aws/apis/aws_provider_spec.go
shreyas-s-rao: If you're sure this is not being consumed by gardener or any other community members who may directly be using MCM, then sure. But I would prefer not doing that in this PR due to scope creep, unless it's absolutely necessary to keep the `MarketType` API consistent.

pkg/aws/apis/aws_provider_spec.go
shreyas-s-rao: Validation is still technically a different package, although inside the same parent directory. If you would still want to use `ec2types.NetworkInterfaceType` instead of string, then I can do it, but would advise against it to keep the API itself lean.

> As for the question of having a separate API module for MCM, I would be interested in understanding the use-case of such a module, is it just for importing the required data for gardener-extension-provider-aws?

I cannot answer this question right now. Few repos in gardener (etcd-druid, gardener) have switched to separate API modules for easier programatic consumption by downstream projects.

pkg/aws/apis/validation/validation.go
shreyas-s-rao: Answered in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247#discussion_r3213077193

pkg/aws/apis/validation/validation.go
shreyas-s-rao: If you have a strong opinion on using `ec2types` in the API, then I can make the change. Please let me know

pkg/aws/core.go
shreyas-s-rao: Thanks, changed in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/changes/077c8e6d83c3f9168f60ff2f6e1e36ca5741b9da

pkg/aws/core_util.go
shreyas-s-rao: Done in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/changes/077c8e6d83c3f9168f60ff2f6e1e36ca5741b9da

shreyas-s-rao:
@takoverflow thanks for your review. I've made some changes and replied to other comments. PTAL, thanks

pkg/aws/apis/aws_provider_spec.go
takoverflow: I consulted with @aaronfern and @elankath and they're of the opinion of keeping the api package lean 👍 

Resolving the comments related to this. Thanks for bearing with my comments.

pkg/aws/apis/aws_provider_spec.go
takoverflow: Consulted with @aaronfern and he confirmed that the support for spot instances is not there in gardener, so we can introduce a slice of market types which would support specifying "spot" or "capacity-block". And based on the type specified, the field
`inputConfig.InstanceMarketOptions` can be appropriately set.

The release note can be updated to notify the breaking change of removal of the `SpotPrice` field.

shreyas-s-rao: Thanks for the inputs @takoverflow @aaronfern . I've made the change in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/changes/59705b2b0ae0137acc95b20f9286cbe3852a6972. PTAL, thanks

shreyas-s-rao:
@takoverflow I've made the requested changes in https://github.com/gardener/machine-controller-manager-provider-aws/pull/247/commits/59705b2b0ae0137acc95b20f9286cbe3852a6972 and also updated the release notes. Please PTAL, thanks.

takoverflow (commented):
/lgtm

Thanks for the PR! And working on enabling support for `capacity-block-reservation` in gardener.

Just one question, there's similar functionality offered by other cloud providers as well, for example: [Future reservations in GCP](https://docs.cloud.google.com/compute/docs/instances/choose-reservation-type#future-reservations)

Are there any plans for adding support for other CSPs as well?

shreyas-s-rao:
Thanks @takoverflow . My first focus was on supporting EFA network interfaces for AWS, which happened to touch upon other concepts like capacity reservations since the more performant GPU instances (NVIDIA H200, B200, etc) mandatorily require reserving capacity beforehand. And I have also worked with Future Reservations in GCP for these larger GPU instances, so I will test it out further before raising the necessary PR(s). But for now, I've only thoroughly tested AWS.

elankath (approved):
/lgtm

PR #251, opened 2026-06-26
Honor machine.Spec.MachineConfiguration.MachineCreationTimeout

elankath:
See https://github.com/gardener/dependency-watchdog/issues/201
and https://github.com/gardener/machine-controller-manager/issues/1098 for background

Providers are not honoring real Machine creation timeout but have hardcoded values to a few minutes. This causes repeated `CrashLoopbackoff` failures when provider is slow.

Fixes #

**Release note**:

```bugfix operator
Honor machine.Spec.MachineConfiguration.MachineCreationTimeout for AWS machine creation instead of using a hardcoded timeout
```

### DIFF
--- a/pkg/aws/core.go
+++ b/pkg/aws/core.go
@@ -54,7 +54,7 @@ const (
 	awsPlacement     = "machine.sapcloud.io/awsPlacement"
 )
 
-var maxElapsedTimeInBackoff = 5 * time.Minute
+var defaultMaxElapsedTimeInBackoff = 5 * time.Minute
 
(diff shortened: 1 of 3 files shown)


takoverflow (commented):
/lgtm

Nice catch finding out about these hardcoded provisioning timeout values

aaronfern:
Looks good!
Can you please elaborate on the release note? something like `Honor machine.Spec.MachineConfiguration.MachineCreationTimeout for AWS machine creation instead of using a hardcoded timeout`

Also, is this a bugfix? if that's the case then release note category should be `bugfix`

/lgtm

PR #255, opened 2026-08-13
Fix multi-arch container image build to compile `TARGETARCH` binary instead of always `amd64`

vpnachev:
Fix multi-arch container image build to compile `TARGETARCH` binary instead of always `amd64`

Fixes #

**Release note**:

```bugfix operator
Fix multi-arch container image build to compile the binary for the configured `TARGETARCH` instead of always for `amd64`, i.e. the `arm64` image now contains `arm64` binary instead of `amd64`. 
```

cc @adenitiu

### DIFF
--- a/.ci/build
+++ b/.ci/build
@@ -48,9 +48,9 @@ fi
 ###############################################################################
 
 # If no LOCAL_BUILD environment variable is set, we configure the `go build` command
-# to build for linux OS, amd64 architectures and without CGO enablement.
+# to build for the local/default OS and ARCH and without CGO enablement.
 if [[ -z "$LOCAL_BUILD" ]]; then
-  CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+  CGO_ENABLED=0 GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) go build \
     -a \
     -v \
     -o ${BINARY_PATH}/rel/machine-controller \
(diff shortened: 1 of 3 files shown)


.ci/build
```diff
@@ -48,9 +48,9 @@ fi
 ###############################################################################
 
 # If no LOCAL_BUILD environment variable is set, we configure the `go build` command
-# to build for linux OS, amd64 architectures and without CGO enablement.
+# to build for the local/default OS and ARCH and without CGO enablement.
 if [[ -z "$LOCAL_BUILD" ]]; then
-  CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+  CGO_ENABLED=0 GOOS=$(go env GOOS) GOARCH=$(go env GOARCH) go build \
```
elankath, line 53: Just a minor question: the expectation is that binaries and `Dockerfile` will always be produced  for native architecture of the machine executing the build. ie we will not support cross compilation - this is what is expected, right ?

elankath (approved):
/lgtm thanks for this - just had a question on expectations.

.ci/build
vpnachev: By default the build is for the host os and architecture, however go supports cross compilation and with this PR it is very easy to achieve it, just run 
```bash
make docker-image TARGET_PLATFORMS=<os/arch>
```
, you can even build multiplatform images locally with 
```bash
make docker-image TARGET_PLATFORMS=linux/amd64,linux/arm64
```

PR #256, opened 2026-08-19
Remove unused CI file

aaronfern:
This PR removed the `pipeline_integration_test` CI file that was unused.
All references to this were also removed

Fixes #

This file exists in the other extension repos as well. I will raise PRs for those once this is merged

**Release note**:

```other developer
Remove unused `.ci/pipeline_integration_test` file
```


elankath (approved):

Makefile
```diff
@@ -77,7 +76,7 @@ update-dependencies:
 
 .PHONY: test-unit
 test-unit:
-	@SKIP_INTEGRATION_TESTS=X .ci/test
+	.ci/test
```
takoverflow, line 79: Now there's no difference between `test-unit` and `test` make targets. I guess one of them can be removed.

aaronfern: Addressed via https://github.com/gardener/machine-controller-manager-provider-aws/pull/256/commits/f6bf517f973cf8b7a9586467d124170c2532f029
I've removed the `test` target, so now we're left with `test-unit` and `test-integration` targets

takoverflow: The `test` target was also used by `verify`, so that should be updated to use `test-unit` as well.

aaronfern: Thanks for catching this!
Updated

takoverflow (commented):
/lgtm

